PR #210, opened 2012-01-16
Add Thor#hide_if

nicknovitski:
hide_if accepts a block, which it stores in the next-defined task, which will evaluate it whenever #hidden? is called, allowing tasks which hide and show themselves dynamically depending on arbitrary conditions.  For example, a cleanup script might only appear in the command list if there were any files that needed cleaning up.

Since Tasks can now have varied and dynamic values for #hidden, the class HiddenTask was removed as unnecessary.

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -66,6 +66,12 @@ class Thor
       end
     end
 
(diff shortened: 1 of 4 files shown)


rafaelfranca:
Thank you for the pull request. Closing due to inactivity.

PR #327, opened 2013-04-07
Refactored duplicate diff methods in color.rb and html.rb

doolin:
Noticed some duplicated code so refactored it. Passes all specs in spec/shell. Naming could be better, willing to rework that.

### DIFF
--- a/lib/thor/shell/color.rb
+++ b/lib/thor/shell/color.rb
@@ -1,11 +1,15 @@
 require 'thor/shell/basic'
+require 'thor/shell/diff_lines'
 
 class Thor
   module Shell
+
     # Inherit from Thor::Shell::Basic and add set_color behavior. Check
     # Thor::Shell::Basic to see all available methods.
     #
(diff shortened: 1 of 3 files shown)


lib/thor/shell/color.rb
```diff
@@ -102,45 +106,18 @@ def can_display_colors?
...
-            actual  = File.binread(destination).to_s.split("\n")
-            content = content.to_s.split("\n")
-
-            Diff::LCS.sdiff(actual, content).each do |diff|
-              output_diff_line(diff)
-            end
-          else
-            super
-          end
+          show_diff_common(destination, content)
```
rafaelfranca: :+1:

doolin:
I'll take a look at this in a day or two, get it rebased, etc. Thanks for taking a look!

rafaelfranca:
Merged in f3d5d19. Thanks

doolin:
Excellent, thank you.

Issue #306, opened 2013-02-16
default_task breaks subcommand help


aspiers:
Using [this test-case](https://gist.github.com/aspiers/1e2943aa27adbbc1ebb8), thor does not output useful help for the subcommand:

```
$ ruby -Ilib test.rb sub help                                 
Usage:
  test.rb help [COMMAND]

Describe subcommands or one specific subcommand
```

If I comment out the `default_task` line, it works correctly:

```
$ ruby -Ilib test.rb sub help      
Tasks:
  test.rb sub first           # This is the first task in the subcommand
  test.rb sub help [COMMAND]  # Describe subcommands or one specific subcommand
  test.rb sub second          # This is the second task in the subcommand
```

Issue #288, opened 2012-12-11
class name


mjobin-mdsol:
I am having an issue with class name as subcommand that are different than the subcommand itself.

I have my code like this...

``` ruby
  class AppCLI < Thor
  end

  class CLI < Thor
    desc "app subcommand", "Manage the apps ...."
    subcommand "app", AppCLI 
  end

```

in my Task list i can see it is registered as "app" properly...

```
Tasks:
  program app subcommand     # Manage the apps ...
```

but when I see the subcommand list, app is shown as app_c_l_i

```
$ program app 
Tasks:
  program app_c_l_i new APP_NAME      # Creates an app...
```

is there an option or setting to override that ? or is it a bug?

thanks

Issue #261, opened 2012-10-10
Default help task for subcommand shows the wrong command name


addiedx44:
I have an executable, ziprecruiter, and it has a subcommand (group?) jobalerts. When I execute ziprecruiter, it displays jobalerts correctly (no underscore):

```
$ bundle exec bin/ziprecruiter
Tasks:
  ziprecruiter help [TASK]          # Describe available tasks or one specific task
  ziprecruiter jobalerts [COMMAND]  # Type ziprecruiter jobalerts for more help.

Options:
  -k, [--api-key=API_KEY]  # Specify your ZipRecruiter Job Alerts API key.
```

However, when I run jobalerts, it shows "c_l_i" instead of the correct "jobalerts":

```
$ bundle exec bin/ziprecruiter jobalerts
Tasks:
  ziprecruiter c_l_i help [COMMAND]      # Describe subcommands or one specific subcommand
  ziprecruiter c_l_i status [TASK_ID]    # A Status action returns the current status of a previously-submitted request.
  ziprecruiter c_l_i subscribe [PATH]    # A Subscribe action is used to upload a collection of job seekers to subscribe to the ZipRecruiter job alerts program.
  ziprecruiter c_l_i unsubscribe [PATH]  # An Unsubscribe action is used to upload a collection of job seekers to unsubscribe from the ZipRecruiter job alerts program.
```

But c_l_i doesn't exist:

```
$ bundle exec bin/ziprecruiter c_l_i
Could not find task "c_l_i".
```

I believe this has to do with the way I have my subcommand namespaced: ZipRecruiter::JobAlerts::CLI

mattheworiordan:
Defining `self.banner` above solved my issues of subcommands either not showing or showing twice with Thor, thank you very much.  Here is an example to help anyone else.

```
#!/usr/bin/env ruby

require 'thor'

class SubCommandBase < Thor
  def self.banner(command, namespace = nil, subcommand = false)
    "#{basename} #{subcommand_prefix} #{command.usage}"
  end

  def self.subcommand_prefix
    self.name.gsub(%r{.*::}, '').gsub(%r{^[A-Z]}) { |match| match[0].downcase }.gsub(%r{[A-Z]}) { |match| "-#{match[0].downcase}" }
  end
end

module App
  class Docs < SubCommandBase
    desc "create", "create docs"
    def create
      # pubish
    end

    desc "publish", "publish docs"
    def publish
      # pubish
    end
  end

  class CLI < Thor
[...]

Issue #170, opened 2011-09-19
Running help on subcommands doesn't show subcommand name


wagenet:
For example `cli subcommand help` shows

```
Tasks:
  cli command1
  cli command2
```

What I expect is:

```
Task:
  cli subcommand command1
  cli subcommand command2
```

This issue also applies to the error message when a subcommand's command is called incorrectly.

jean-francois-labbe:
+1

albrechtsimon:
+1

PR #330, opened 2013-05-09
Possible fix for the subcommand help messages using the klass name instead of command name

rberger:
Changed spec/subcommand and spec/fixtures/script.thor to expose the problem of when the subcommand name is different than the klass.

Fixed it by:

Updating lib/thor.rb metaprogramming magic that adds the help methods
to the subcommand klass to include a class instance variable that is
initialized to the subcommand_name and a method to access it.

Then modified command#formatted_usage to make use of that method if
its processing a subcommand and the subcommand_method exists.

Also had to "fix" the spec/register_spec.rb
   context "when $thor_runner is false"
that had a test that seemed to accecpt the fact that subcommand names
got foobar'd if the klass name is different than the subcommand name.
Or I didn't understand that test, but I changed it to what made more
sense to me.

I believe this addresses #306, #169, #170, #196, #261, #128, #288

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -454,7 +454,11 @@ class Thor
 
     def subcommand_help(cmd)
       desc "help [COMMAND]", "Describe subcommands or one specific subcommand"
-      class_eval <<-RUBY
+      class_eval <<-RUBY 
+        @subcommand_name = "#{cmd.to_s}"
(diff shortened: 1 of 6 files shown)


rafaelfranca:
Closing since this is too outdated. Thank you for the pull request.

PR #341, opened 2013-06-20
Smart directory

foxnewsnetwork:
TL;DR: added Actions#smart_directory which takes the same arguments as #directory, but allows users to specify a ".zc" file to complement a ".tt" file such that content in the .zc file is rendered then inserted the ".tt" file in sections marked "### Insert here ###" so that re-invocations will inject code when it would otherwise overwrite.

Motivation: I am a web app code slave and, for my work, I spend all day writing extremely similar code for a Ruby on Rails shop. To keep me sane, I built a code scaffolder that uses thor to copy over fairly complicated rails resources setups (for example, suppose ":" represents a has_many relationship, and "+" represents no relation, I built something to rapidly scaffold out a bunch of assets "negotiation:(offer:(listing+user)+comment:user+listing:user" ). Unfortunately, using Actions#directory doesn't make use of the state of the destination folder already there, and so calling the #directory methods repeatedly makes thor want to just overwrite files instead of adding onto them.


rafaelfranca:
Closing since this is too outdated and we don't have interest in the feature. Thank you for the pull request

PR #412, opened 2014-04-19
Use custom print method for printing defaults.

vStone:
Prints default arrays in a copy paste able way.

Difference in output:

Old:

```
Options:
  [--ignore-files=one two three]
                                       # Default: ["**/spec/**/*.pp", "**/tests/**/*.pp", "**/example/**/*.pp"]
  [--disable-manifests=one two three]
                                       # Default: autoloader_layout
```

New:

```
Options:
  [--ignore-files=one two three]
                                        # Default: "**/spec/**/*.pp" "**/tests/**/*.pp" "**/example/**/*.pp"
  [--disable-manifests=one two three]
                                       # Default: autoloader_layout
  [--verbose], [--no-verbose]
```

### DIFF
--- a/lib/thor/base.rb
+++ b/lib/thor/base.rb
@@ -518,7 +518,7 @@ class Thor
             item.push(option.description ? "# #{option.description}" : "")
 
             list << item
-            list << ["", "# Default: #{option.default}"] if option.show_default?
+            list << ["", "# Default: #{option.print_default}"] if option.show_default?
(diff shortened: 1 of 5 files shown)


rafaelfranca:
Merged in [2ea25a6](https://github.com/rails/thor/commit/2ea25a61dad29132ad48a192dfa3b104e4007725)

PR #425, opened 2014-06-27
Allow looking inside engines folder

ikusuki:
Instead of looking just in one place for thor files, I propose searching also inside engines folder and grab all of them together.

This way, we can split thor tasks for each engine and have these for main application as well in the same place they were before.

### DIFF
--- a/lib/thor/runner.rb
+++ b/lib/thor/runner.rb
@@ -242,11 +242,13 @@ private
 
     unless skip_lookup
       Pathname.pwd.ascend do |path|
-        thorfiles = Thor::Util.globs_for(path).map { |g| Dir[g] }.flatten
-        break unless thorfiles.empty?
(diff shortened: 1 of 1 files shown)


lib/thor/runner.rb
```diff
@@ -242,11 +242,13 @@ def thorfiles(relevant_to = nil, skip_lookup = false)
 
     unless skip_lookup
       Pathname.pwd.ascend do |path|
-        thorfiles = Thor::Util.globs_for(path).map { |g| Dir[g] }.flatten
-        break unless thorfiles.empty?
+        thorfiles.push Thor::Util.globs_for(path).map { |g| Dir[g] }.flatten
       end
+      Pathname.new("./engines").children.each do |path|
```
sferik: This line is causing tests to fail with the error:

```
No such file or directory - ./engines (Errno::ENOENT)
```

PR #426, opened 2014-07-11
Fix for an array of colors with ask/yes?/no?

tjwp-yesware:
This fixes a problem using an array of colors or the
legacy color syntax with `ask` or any of the methods
that depend on it.

The symptom of the problem is that the color arg would
prefix the message:

    [:green, :on_red, :bold]Not what you expected?
    [:blue, false]Why isn't this blue?


sferik:
Thanks for the fix!

PR #433, opened 2014-08-23
ask_filtered should allow users a skip option if they do not want to sel...

karimainine:
adding a skip option to ask_filtered with a description label before the :limited values


rafaelfranca:
Closing since there was no update.

PR #434, opened 2014-08-25
First steps to provide localization for thor

ghost:
Hi there,

I use`thor` in my projects a lot and would like to provide the users a localized cli. Parts are done, but messages generated by `thor` itself are missing.

This is a very rough version, which I send to you to get some feedback. I would like to add a very thin layer to `thor` just to make it possible to localize its messages. The hard work should be done outside `thor`, e.g. by `I18n` or some other library.

For now I changed the messages in `lib/thor.rb` to give you an idea about the API. I took some code from `I18n` to make interpolation possible. Therefor it has the same functionality like `I18n` in regards of string interpolation.

In this version I use a `YAML`-file to store the translations. But this is just for my convenience during development. To speed things up, I can put that stuff into a `Hash` or some similar data structures. 

Would you mind to give me some feedback? Would you merge such an PR if it matches your quality requirements etc..

Thank you very much.

Regards,
dg

### DIFF
--- a/lib/thor/localize.rb
+++ b/lib/thor/localize.rb
@@ -1,4 +1,3 @@
-require 'psych'
 require "thor/localize/internal"
 require "thor/localize/interpolate"
 
--- a/lib/thor/localize/internal.rb
+++ b/lib/thor/localize/internal.rb
@@ -1,10 +1,12 @@
+require 'yaml'
+
 class Thor
   module Localize
     module Internal
       class << self
         def t(translation_key, interpolation_values)
           file_path = File.expand_path('../internal.yaml', __FILE__)
-          data = Psych.load(File.read(file_path))
+          data = YAML.load(File.read(file_path))
 
           string = translation_key.split(/\./).inject(data) do |a, e|
             fail InvalidLocalizationKeyError, "Translation missing: #{translation_key}" unless a.key?(e)


dg-ratiodata:
As mentioned from @wycats I will add some more explanation to this PR. @wycats Can you give me some feedback on this? Thanks.

### Reason for change

I'm using `thor` in most of my projects and want to give users a localized commandline interface. For now there are a lot of English messages hidden in `thor`'s source code.

### Proposal

This PR will replace all strings output by `puts` et al. by using a syntax similar to `I18n`.

```ruby
puts "I'm a quite long message"
```

will become something like

```ruby
puts Thor::Localize.t('thor.help.usage')
```

The api I have in mind:
1. Is a thin wrapper
2. By default uses a small backend
3. The default backend can be substituted a more advance one like I18n.

To make 1. possible I will add something like that

```ruby
class Thor
  module Localize
    @backend = Thor::Localize::Internal

    attr_accessor :backend

    def t(translation_key, interpolation_values = {})
      backend.t(translation_key, interpolation_values)
    end

    module_function :backend, :t
  end
end
```


The default backend uses `YAML` in the background. Before that I tried several other approaches like using a `Module`, but this was hard to maintain. After playing around which `YAML`-approach a bit. I found it quite handy. Translations are easy to change/add later on. The structure is clear and lean - no need for `{}` (Hash) etc. `YAML` is part of `ruby-core`. As `I18n` also uses `YAML` files and you can reuse the default [ file ]( https://github.com/dg-ratiodata/thor/blob/feature/localize_thor_help_2/lib/thor/localize/internal.yaml ) and only change the translations where needed.
 
All translations are placed in a separate `YAML`-file in the [ `localize`-subdirectory ]( https://github.com/dg-ratiodata/thor/tree/feature/localize_thor_help_2/lib/thor/localize ). The default backend also supports interpolation. The code for that was [ "borrowed" ]( https://github.com/svenfuchs/i18n/blob/master/lib/i18n/interpolate/ruby.rb ) from `I18n` - which is also licensed under the [ MIT-license ]( https://github.com/svenfuchs/i18n/blob/master/MIT-LICENSE ).

To support much more complex setups, one can exchange (3.) the backend by simply using `Thor::Localize.backend = NewBackend`. The only requirement for a backend would be, that it needs to support `#t`. Because all "requests" will be forwarded to that method.

Things where I need some feedback / which need to be discussed

* [ ] Is using `YAML` ok for you? I don't think that using it should have any negative side affects on the performance
* [ ] Should exchanging backends be supported?
* [ ] The structure of messages in the `YAML`-file

   I think this is a tough one. For now I'm using this [ structure ]( https://github.com/dg-ratiodata/thor/blob/feature/localize_thor_help_2/lib/thor/localize/internal.yaml ). Every thor commandline command - `update`, `list_installed` etc. - has it's own section. I would try to make use of that structure everywhere. BTW there's an identation error [ here ]( https://github.com/erikhuda/thor/pull/434/files#diff-792c19d22c8092d7717d6a7ee8fd8a59R23 ) that would be fixed in the finished PR.

* [ ] Testing
  
  What would you want me to test? From my side I would add tests for the default backend, the interpolation-feature, exchanging the backend and some messages.

I would be grateful to get some feedback. Thanks a lot.

wycats:
@dg-ratiodata have you considered making the API compatible with the Rails localization interface? There are several useful backends that already exist for the Rails interface that might be handy here.

dg-ratiodata:
Mmh.. I'm not sure what you mean  with "API compatible". I thought Rails uses I18n. Shouldn't that mean you've got everything in place when you use `I18n` as backend?

```ruby
Thor::Localize.backend = I18n
```

dg-ratiodata:
@wycats Is this what you have in mind?

maxmeyer:
Friendly ping @wycats Anything I missed here? The idea was to provide a minimal layer which makes it possible to use fullblown solutions like I18n or something selfmade. I don't want `thor` to require I18n to be installed.

dg-ratiodata:
Friendly ping @wycats Anything I missed here? The idea was to provide a minimal layer which makes it possible to use fullblown solutions like I18n or something selfmade. I don't want thor to require I18n to be installed.

PR #435, opened 2014-08-27
Check the type of the default

huwr:
Hi,

I think the default option should match the type.  This is a problem I ran into while using Thor.  I had code like this:

``` ruby
method_option :foo, :aliases => '-f', :default => '50', :type => :numeric
```

I expect the default value `'50'` to be a numeric, but of course it was a string. 

Hopefully, this pull request solves the problem. ;-)

Cheers,
Huw.

### DIFF
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -108,6 +108,21 @@ class Thor
 
     def validate!
       fail ArgumentError, "An option cannot be boolean and required." if boolean? && required?
+      validate_default_type!
+    end
+
+    def validate_default_type!
+      default_type = case @default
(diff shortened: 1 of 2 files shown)


rafaelfranca:
I believe this will be a breaking change to some project. While they can forward fix the mismatch in the type the current releases of those project will now start to raise errors when loading the generator. This was the case of rails https://github.com/rails/rails/pull/22789.

I agree this is a good change but I think we should release it in a non-backward compatible release. Maybe a warning instead of an exception for the current release?

robdimarco:
@rafaelfranca comment above is spot on and is the cause of issue #533

PR #436, opened 2014-08-29
Add support for `HashWithIndifferentAccess#key?` and its aliases

jfarmer:
This commit adds indifferent versions of `Hash#key?` and its aliases (`Hash#has_key?`, `Hash#include?`, and `Hash#member?`) to `Thor::CoreExt::HashWithIndifferentAccess`.

It works thus:

``` ruby
hash = Thor::CoreExt::HashWithIndifferentAccess.new('foo' => 'bar')
hash.key?('foo') # => true
hash.key?(:foo)  # => true
```

Before this commit, `hash.key?(:foo)` would have returned `false`.

### DIFF
--- a/lib/thor/core_ext/hash_with_indifferent_access.rb
+++ b/lib/thor/core_ext/hash_with_indifferent_access.rb
@@ -24,6 +24,14 @@ class Thor
         super(convert_key(key), value)
       end
 
+      def key?(key)
+        super(convert_key(key))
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Why do you need this change? `Thor::CoreExt::HashWithIndifferentAccess` is an internal class and should not be used by users. How are you using it?

PR #446, opened 2014-10-14
Don't freeze the options hash

mmay:
I couldn't figure out why freezing the options hash was a good idea. Any reason why the options hash is frozen?

In my case, I want to override an option after it has already been set and instead of using an instance variable I would prefer to just override the options hash.

### DIFF
--- a/lib/thor/parser/options.rb
+++ b/lib/thor/parser/options.rb
@@ -111,9 +111,7 @@ class Thor
 
       check_requirement!
 
-      assigns = Thor::CoreExt::HashWithIndifferentAccess.new(@assigns)
-      assigns.freeze
-      assigns
+      Thor::CoreExt::HashWithIndifferentAccess.new(@assigns)
     end
 
     def check_unknown!


sferik:
It appears that the call to `freeze` was added in https://github.com/erikhuda/thor/commit/f7d9cc61cc20809f4dd050c49a33fdd9a2e4ebad.

@josevalim Do you remember why this was necessary?

josevalim:
It was frozen before here: https://github.com/erikhuda/thor/commit/f7d9cc61cc20809f4dd050c49a33fdd9a2e4ebad#diff-0065e6d4e143036f62fc86861719ddc1L117

It any case, it is probably due to safety. @mmay I think that instead of overriding it after initializing it, you should provide a class method that will build it with the proper option set in the first place?

mmay:
Thanks for taking a look @sferik and @josevalim!

Using a class method to build the options hash is a better way to do what I wanted anyway.

Issue #455, opened 2014-11-19
:hash type should validate keys are specified *at most once*


dgorst:
Hi folks - what's the consensus on how Thor should handle duplicate keys in hash arguments?

Here's a dumb example:

```
      desc 'test','Just play around with hash arguments'
      method_option :data, :required => true, :type => :hash, :aliases => '-d', :desc => 'Some hash'
      def test
        puts options[:data]
      end
```

You run that on the command line specifying the same hash key multiple times, and you get this:

```
pingpong:~ dgorst$ play test -d bar:fu bar:foo
{"bar"=>"foo"}
```

Now, obviously the Hash itself can only be expected to contain one value for a key, but I'd expect Thor to bork validating what the user has typed (or at least give me the option to make it complain to the user).

It's nice to give users a bit of help and fast feedback when they are doing something wrong, and I think this would be helpful (well, I know it would have been with an issue we've had recently).

Any objections or comments on this behaviour?

PR #456, opened 2014-11-20
Issue #455 - only allow a hash key to be specified once

dgorst:
Just a small update to validate a user only enters a key for a hash option at most once. Issue #455


sferik:
Thanks!

PR #460, opened 2014-12-02
Negative options

aderouineau:
Do not generate a negative option (--no-no-foo) for already negative boolean options (--no-foo)

### DIFF
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -86,7 +86,7 @@ class Thor
       sample = "[#{sample}]" unless required?
 
       if boolean?
-        sample << ", [#{dasherize("no-" + human_name)}]" unless name == "force"
+        sample << ", [#{dasherize("no-" + human_name)}]" unless name == "force" or name.start_with?("no-")
       end
 
       if aliases.empty?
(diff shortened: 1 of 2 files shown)


Fudoshiki:
+1  An option's default must match its type  error in master

lib/thor/parser/option.rb
```diff
@@ -86,7 +86,7 @@ def usage(padding = 0)
       sample = "[#{sample}]" unless required?
 
       if boolean?
-        sample << ", [#{dasherize("no-" + human_name)}]" unless name == "force"
+        sample << ", [#{dasherize("no-" + human_name)}]" unless name == "force" or name.start_with?("no")
```
sferik: I believe this line would prevent the automatic negation of all options that that start with `"no"`, even if they are not negative options?

For example, what if there was an boolean `--nougat` option. I believe this would prevent the generation of a `--no-nougat` option.

aderouineau: Good catch. I fixed it and also added a test case for "--nougat" to make sure there will never be a regression :)

sferik:
Thanks!

PR #470, opened 2015-01-21
Add #fetch to HashWithIndifferentAccess

justincampbell:
I ran into surprising behavior when `options[:foo] || 123` worked as expected, but switching to `options.fetch(:foo, 123)` always resulted in `123`.

I know that defaults should be set in option/method_option, but the class that I'm instantiating is used by other clients than Thor.


sferik:
Closed by https://github.com/erikhuda/thor/commit/f0c2166534e122636f5ce04d61885736ef605617.

Issue #417, opened 2014-05-12
Allow suppression of auto-created 'no-' flags for boolean options


mbrgm:
The name of my option is `--all`. But I do not want the usage instructions to include a `--no-all` flag, as this is already the default behaviour.

https://github.com/erikhuda/thor/blob/d634d240bdc0462fe677031e1dc6ed656e54f27e/lib/thor/parser/option.rb#L89 already excludes the automatically inverted flag when the option name is 'force'. Maybe one could simply turn this into a metadata option (and include 'force' in some kind of blacklist for auto-inversed options).

``` gherkin
Feature: Automatically created '--no-' options can be suppressed
  As a developer
  I want to be able to suppress the automatically created '--no-' flag for boolean options
  In order to remove them from the usage description

  Scenario: Automatically created '--no-' flag is suppressed explicitely
    When I create an option `:foo`
    And I set the option's `:type` to `:boolean`
    And I supply a metadata key `:auto_inversion` with a value of `false`
    Then the usage instructions of the command should not include a '--no-foo' flag

  Scenario: Automatically created '--no-' flag is requested explicitely
    When I create an option `:foo`
    And I set the option's `:type` to `:boolean`
    And I supply a metadata key `:auto_inversion` with a value of `true`
    Then the usage instructions of the command should include a '--no-foo' flag

  Scenario: Automatically created '--no-' flag is created by default
    When I create an option `:foo`
    And I set the option's `:type` to `:boolean`
    Then the usage instructions of the command should include a '--no-foo' flag
```

ddarbyson:
@afeld agreed. Have you found any way to acheive this?

afeld:
@ddarbyson I hadn't, and typed out a response saying so...but figured "what the hell", so took a couple passes at it last night (who needs sleep anyway??) Would love feedback on the PRs linked above.

ddarbyson:
@afeld I tried your patch, works just fine. I ended up using it in the following fashion to have cleaner command line usage information when calling `--help`

```ruby
    method_option '--no-remove', :type => :boolean, :desc => 'Do not remove contents before clone', :default => false
    method_option '--no-shallow', :type => :boolean, :desc => 'Do not truncate history during clone', :default => false
```

Then somewhere below, always default to 'false' unless `--no-shallow` is passed as a 'true' option
```ruby
# this will always run by default as specified by method_option :default => false
unless option['no-shallow']
  ...
end
```

[...]

PR #473, opened 2015-01-30
deprecate generation of inverse flags for booleans

afeld:
To accommodate #417, this PR removes the documentation of the inverse for boolean flags, e.g. `--no-foo` for `--foo`, and vice versa. Note that the functionality is still supported under the hood.

### DIFF
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -85,10 +85,6 @@ class Thor
 
       sample = "[#{sample}]" unless required?
 
-      if boolean?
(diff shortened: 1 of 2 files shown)


afeld:
Bump!

rafaelfranca:
I don't think we can disable this by default, but we can accommodate a option to disable it in the DSL. Can you provide a PR doing that? Thanks

xjunior:
Opened this to allow the suppression of the inverse flags: https://github.com/erikhuda/thor/pull/733

PR #475, opened 2015-01-30
remove special handling of inverse boolean flags

afeld:
:warning: **WIP** :warning: 

Closes https://github.com/erikhuda/thor/issues/417#issuecomment-45052020.

This is a superset of #473, removing all special handling of boolean flags generating/parsing the inverse. It certainly makes the behavior more straightforward (no magic flags), though I realize checks for "negative" flags (like `options[:no_public] == true`) become a little more confusing. Thoughts? Also, I'm new to this codebase, so while all tests are passing, would love another set of :eyes: to ensure I got everything.

### DIFF
--- a/lib/thor/parser/arguments.rb
+++ b/lib/thor/parser/arguments.rb
@@ -55,11 +55,6 @@ class Thor
 
   private
 
-    def no_or_skip?(arg)
-      arg =~ /^--(no|skip)-([-\w]+)$/
-      $2
-    end
-
     def last?
       @pile.empty?
     end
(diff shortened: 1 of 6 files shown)


fny:
I just ran into this when wrapping an AWS CLI command that had a `--no-verify-ssl` option. Thor ended up spitting out `--no-verify-ssl` and `--no-no-verify-ssl`.

In certain cases, it would be nice if there was a way to get both through a specified type (e.g. `:magic_boolean`).

rafaelfranca:
See https://github.com/erikhuda/thor/pull/473#issuecomment-279279157

PR #476, opened 2015-02-02
Fix subcommand again

akm:
I fixed the same bug as groovenauts-thor-0.19.1 in https://github.com/groovenauts/thor/pull/1 but that PR was made from the HEAD of master branch. So it includes other changes which modify behavior `rails generate`.

This is the same fix but made from v0.19.1 tag. And this will be tagged groovenauts-thor-0.19.1.1


akm:
Sorry. This is a mistake.

PR #478, opened 2015-02-05
Add key? method for improving ::Hash compatibility

kiyohara:
options (HashWithIndifferentAccess) has key checkable by either strings or symbols


sferik:
Thanks!

PR #483, opened 2015-04-01
Add a new feature for exclusive options and required at_least_one option

nstoym:
🌈
I've added a new feature for exclusive options and required at least one option.
For exclusive options,

``` ruby
exclusive do
  option :one
  option :two
end
```

You cann't give option `--one` and `--two` at the same time.
If you give wrong options, it shows an error as follows:

```
Found exclusive options '--one`, `--two`
```

For required at least one option:

``` ruby
at_least_one do
  option :three
  option :four
end
```

You must give  at least one option `--three` or `--four`.
This error message:

```
Not found at least one of required options `--three`, `--four`
```

`#class_exclusive` and `#class_at_least_one` methods can be used for `class_option`.
If `#excusive`/`#class_exclusive` or `#at_least_one`/`#class_at_least_one` are defined, the help command show you the relation of options after the`Options:` list as follows:

```
Exclusive Options:
  --one   --two

Required At Least One:
  --three  --four
```

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -160,6 +160,97 @@ class Thor # rubocop:disable ClassLength
       @disable_class_options = true
     end
 
+    # Returns this class exclusive options array set.
+    #
+    # ==== Rturns
+    # Array[Array[Thor::Option.name]]
+    #
+    def method_exclusive_option_names
+      @method_exclusive_option_names ||=[]
+    end
+
+    # Adds and declareds option group for exclusive options in the
+    # block and arguments. You can declare options as the outside of the block.
+    # If :for is given as option,
+    # it allows you to change the options from a prvious defined command.
+    #
+    # ==== Parameters
+    # Array[Thor::Option.name]
+    # options<Hash>:: :for is applied for previous defined command.
+    #
+    # ==== Examples
+    #
+    #   exclusive do
+    #     option :one
+    #     option :two
+    #    end
+    #
+    # Or
+    #
+    #   option :one
+    #   option :two
+    #   exclusive :one, :two
+    #
+    # If you give "--one" and "--two" at the same time.
+    # ExclusiveArgumentsError will be raised.
+    #
+    def method_exclusive(*args, &block)
+      register_options_relation_for(:method_options,
+                                    :method_exclusive_option_names, *args, &block)
+    end
+    alias_method :exclusive, :method_exclusive
+
+    # Returns this class at least one of required options array set.
+    #
+    # ==== Rturns
+    # Array[Array[Thor::Option.name]]
+    #
+    def method_at_least_one_option_names
+      @method_at_least_one_option_names ||=[]
+    end
+
+    # Adds and declareds option group for required at least one of options in the
+    # block of arguments. You can declare options as the outside of the block.
+    # If :for is given as option,
+    # it allows you to change the options from a prvious defined command.
+    #
+    # ==== Parameters
+    # Array[Thor::Option.name]
+    # options<Hash>:: :for is applied for previous defined command.
+    #
+    # ==== Examples
+    #
+    #   at_least_one do
+    #     option :one
+    #     option :two
+    #    end
+    #
+    # Or
+    #
+    #   option :one
+    #   option :two
+    #   at_least_one :one, :two
+    #
+    # If you do not give "--one" and "--two".
+    # AtLeastOneRequiredArgumentError will be raised.
+    #
+    # You can use at_least_one and exclusive at the same time.
+    #
+    #    exclusive do
+    #      at_least_one do
+    #        option :one
+    #        option :two
+    #      end
+    #    end
+    #
+    # Then it is required either only one of "--one" or "--two".
+    #
+    def method_at_least_one(*args, &block)
+      register_options_relation_for(:method_options,
+                                    :method_at_least_one_option_names, *args, &block)
+    end
+    alias_method :at_least_one, :method_at_least_one
+
     # Prints help information for the given command.
     #
     # ==== Parameters
@@ -175,6 +266,9 @@ class Thor # rubocop:disable ClassLength
       shell.say "  #{banner(command)}"
       shell.say
       class_options_help(shell, nil => command.options.map { |_, o| o })
+      print_exclusive_options(shell, command)
+      print_at_least_one_required_options(shell, command)
+
       if command.long_description
         shell.say "Description:"
         shell.print_wrapped(command.long_description, :indent => 2)
@@ -205,6 +299,8 @@ class Thor # rubocop:disable ClassLength
       shell.print_table(list, :indent => 2, :truncate => true)
       shell.say
       class_options_help(shell)
+      print_exclusive_options(shell)
+      print_at_least_one_required_options(shell)
     end
 
     # Returns commands ready to be printed.
@@ -219,6 +315,26 @@ class Thor # rubocop:disable ClassLength
     end
     alias_method :printable_tasks, :printable_commands
 
+    def print_exclusive_options(shell, command = nil)
+      opts = []
+      opts  = command.method_exclusive_option_names unless command.nil?
+      opts += class_exclusive_option_names
+      unless opts.empty?
+        shell.say "Exclusive Options:"
+        shell.print_table(opts.map{ |ex| ex.map{ |e| "--#{e}"}}, :indent => 2 )
+        shell.say
+      end
+    end
+    def print_at_least_one_required_options(shell, command = nil)
+      opts = []
+      opts = command.method_at_least_one_option_names unless command.nil?
+      opts += class_at_least_one_option_names
+      unless opts.empty?
+        shell.say "Required At Least One:"
+        shell.print_table(opts.map{ |ex| ex.map{ |e| "--#{e}"}}, :indent => 2 )
+        shell.say
+      end
+    end
     def subcommands
       @subcommands ||= from_superclass(:subcommands, [])
     end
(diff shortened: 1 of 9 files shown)


nstoym:
I can not buid test enviroments for Ruby 1.8.7 by `bundle install` for the following install error.

```
Gem::InstallError: celluloid requires Ruby version >= 1.9.2.
An error occurred while installing celluloid (0.16.0), and Bundler cannot continue.
Make sure that `gem install celluloid -v '0.16.0'` succeeds before bundling.
```
So, I need to insert the dependency in Gemfile.

```
platforms :ruby_18 do
  gem 'celluloid', '0.11.0'
end
```
I don't know whether this is correct. If you continue to support Ruby 1.8.7, I think it is necessary to fix this issue.

sferik:
This looks good but it contains merge conflicts. Would you mind pulling from the upstream `master` and rebasing your branch?

nstoym:
Hi sferik!

Thanks for your reply.
I will rebase soon.

> Can you please look into why this doesn’t produce an error message on Ruby 1.8.7?

This error message happened when I built test environment of Ruby 1.8.7 by using command `bundle install`.
"guard-rspec" gem in Gemfile depends on "celluloid" indirectly.
http://gemgraph.herokuapp.com/page?gem_name=guard-rspec&version=

"celluloid" gem requires Ruby >= 1.9.2 from version 0.12.0.
https://github.com/celluloid/celluloid/blob/v0.12.0/celluloid.gemspec

If we need "guard-rspec" gem, we should fix the Gemfile for new developers who don't have environment of Ruby 1.8.7.

sferik:
> If we need "guard-rspec" gem, we should fix the Gemfile for new developers who don't have environment of Ruby 1.8.7.

We definitely don’t need `guard-rspec` in the `Gemfile`. I’ll remove it.

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
       @disable_class_options = true
     end
 
+    # Returns this class exclusive options array set.
+    #
+    # ==== Rturns
+    # Array[Array[Thor::Option.name]]
+    #
+    def method_exclusive_option_names
```
rafaelfranca: This method should be protected and nodoc

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+      @method_exclusive_option_names ||=[]
```
rafaelfranca: Space after the =

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+    end
+
+    # Adds and declareds option group for exclusive options in the
```
rafaelfranca: `declares`

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+    # block and arguments. You can declare options as the outside of the block.
+    # If :for is given as option,
+    # it allows you to change the options from a prvious defined command.
```
rafaelfranca: `previous`

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+    #
+    # ==== Parameters
+    # Array[Thor::Option.name]
+    # options<Hash>:: :for is applied for previous defined command.
+    #
+    # ==== Examples
+    #
+    #   exclusive do
```
rafaelfranca: Could you show an example with the for option?

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+    #     option :two
+    #    end
+    #
+    # Or
+    #
+    #   option :one
+    #   option :two
+    #   exclusive :one, :two
+    #
+    # If you give "--one" and "--two" at the same time.
```
rafaelfranca: No need the `.` in the end.

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+                                    :method_exclusive_option_names, *args, &block)
+    end
+    alias_method :exclusive, :method_exclusive
+
+    # Returns this class at least one of required options array set.
+    #
+    # ==== Rturns
+    # Array[Array[Thor::Option.name]]
+    #
+    def method_at_least_one_option_names
```
rafaelfranca: This method should be protected and nodoc

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+      @method_at_least_one_option_names ||=[]
+    end
+
+    # Adds and declareds option group for required at least one of options in the
```
rafaelfranca: declares

lib/thor.rb
```diff
@@ -160,6 +160,97 @@ def disable_class_options
...
+    #     option :two
+    #    end
+    #
+    # Or
+    #
+    #   option :one
+    #   option :two
+    #   at_least_one :one, :two
+    #
+    # If you do not give "--one" and "--two".
```
rafaelfranca: No need the `.` in the end here

lib/thor.rb
```diff
@@ -219,6 +315,26 @@ def printable_commands(all = true, subcommand = false)
     end
     alias_method :printable_tasks, :printable_commands
 
+    def print_exclusive_options(shell, command = nil)
+      opts = []
+      opts  = command.method_exclusive_option_names unless command.nil?
+      opts += class_exclusive_option_names
+      unless opts.empty?
+        shell.say "Exclusive Options:"
+        shell.print_table(opts.map{ |ex| ex.map{ |e| "--#{e}"}}, :indent => 2 )
```
rafaelfranca: no need the space before the `)`

lib/thor.rb
```diff
@@ -219,6 +315,26 @@ def printable_commands(all = true, subcommand = false)
...
+        shell.say
+      end
+    end
+    def print_at_least_one_required_options(shell, command = nil)
+      opts = []
+      opts = command.method_at_least_one_option_names unless command.nil?
+      opts += class_at_least_one_option_names
+      unless opts.empty?
+        shell.say "Required At Least One:"
+        shell.print_table(opts.map{ |ex| ex.map{ |e| "--#{e}"}}, :indent => 2 )
```
rafaelfranca: No need the space before the `)`

lib/thor/base.rb
```diff
@@ -62,8 +62,20 @@ def initialize(args = [], local_options = {}, config = {}) # rubocop:disable Met
...
       # a list of arguments that weren't declared.
-      stop_on_unknown = self.class.stop_on_unknown_option? config[:current_command]
-      opts = Thor::Options.new(parse_options, hash_options, stop_on_unknown)
+      current_command = config[:current_command]
+      stop_on_unknown = self.class.stop_on_unknown_option? current_command
+
+      # Give a relation of options.
+      # After parsing, Thor::Options check whether right relations are kept
+      relations = {:exclusive_option_names => [], :at_least_one_option_names => []}
+      unless current_command.nil?
```
rafaelfranca: I think here a if/else would be clear

lib/thor/base.rb
```diff
@@ -239,6 +251,24 @@ def arguments
         @arguments ||= from_superclass(:arguments, [])
       end
 
+      # Returns this class exclusive options array set, looking up in the ancestors chain.
+      #
+      # ==== Rturns
+      # Array[Array[Thor::Option.name]]
+      #
+      def class_exclusive_option_names
```
rafaelfranca: Those methods should be protected and nodoc

lib/thor/base.rb
```diff
@@ -274,6 +304,67 @@ def class_option(name, options = {})
         build_option(name, options, class_options)
       end
 
+      # Adds and declareds option group for exclusive options in the
```
rafaelfranca: declares

lib/thor/base.rb
```diff
@@ -274,6 +304,67 @@ def class_option(name, options = {})
...
+      #
+      # If you give "--one" and "--two" at the same time.
+      # ExclusiveArgumentsError will be raised.
+      #
+      def class_exclusive(*args, &block)
+        register_options_relation_for(:class_options,
+                                      :class_exclusive_option_names, *args, &block)
+      end
+
+      # Adds and declareds option group for required at least one of options in the
```
rafaelfranca: declares

lib/thor/base.rb
```diff
@@ -651,6 +742,34 @@ def initialize_added #:nodoc:
       def dispatch(command, given_args, given_opts, config) #:nodoc:
         fail NotImplementedError
       end
+
+      # Register a relation of options for target(method_option/class_option)
+      # by args and block.
+      def register_options_relation_for( target, relation, *args, &block)
```
rafaelfranca: No need the space after the `(`

lib/thor/base.rb
```diff
@@ -651,6 +742,34 @@ def initialize_added #:nodoc:
...
+      # of before and after by block evaluation.
+      def built_option_names(target, opt = {}, &block)
+        before = command_scope_member(target, opt).map{ |k,v| v.name }
+        instance_eval(&block)
+        after  = command_scope_member(target, opt).map{ |k,v| v.name }
+        after - before
+      end
+
+      # Get command scope member by name.
+      def command_scope_member( name, options = {} ) #:nodoc:
```
rafaelfranca: no need the spaces around the ()

lib/thor/base.rb
```diff
@@ -651,6 +742,34 @@ def initialize_added #:nodoc:
...
+        if options[:for]
+          find_and_refresh_command(options[:for]).send(name)
+        else
+          send( name )
```
rafaelfranca: same here about the space

lib/thor/command.rb
```diff
@@ -62,6 +63,12 @@ def formatted_usage(klass, namespace = true, subcommand = false)
       # Strip and go!
       formatted.strip
     end
+    def method_exclusive_option_names
```
rafaelfranca: Missing a blank line above this method and below it

IotaSpencer:
how similar is the code base if i were to update this to the latest thor codebase, as this is one of the features on my wish list for ruby command line frameworks

rafaelfranca:
Merged in [291d93d](https://github.com/rails/thor/commit/291d93d6cb9f81123453f86292974bb7f0621300). Thanks!

PR #498, opened 2015-08-25
Allow values to include hyphens when specified unambiguously with "="

fancyremarker:
:rainbow: 

I noticed that if an option value begins with a "-" (`/^-/`), it will be parsed as a switch, even in cases like

```
--foo=-bar
```

where it seems that the presence of the "=" allows the value (`-bar`) to be unambiguously parsed as a value instead of a switch.

The new spec expectation should be clear; I'm less confident in the implementation, as the current implementation relies quite a bit on several stateful instance variables.

### DIFF
--- a/lib/thor/parser/options.rb
+++ b/lib/thor/parser/options.rb
@@ -72,6 +72,7 @@ class Thor
     def parse(args) # rubocop:disable MethodLength
       @pile = args.dup
       @parsing_options = true
+      @force_value = false
 
       while peek
         if parsing_options?
@@ -83,7 +84,11 @@ class Thor
             when SHORT_SQ_RE
               unshift($1.split("").map { |f| "-#{f}" })
               next
-            when EQ_RE, SHORT_NUM
+            when EQ_RE
+              unshift($2)
+              switch = $1
+              @force_value = true
+            when SHORT_NUM
               unshift($2)
               switch = $1
             when LONG_RE, SHORT_RE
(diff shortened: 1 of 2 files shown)


lib/thor/parser/options.rb
```diff
@@ -129,6 +134,7 @@ def check_unknown!
     # Two booleans are returned.  The first is true if the current value
     # starts with a hyphen; the second is true if it is a registered switch.
     def current_is_switch?
+      return false if @force_value
```
rafaelfranca: This should return an array.

spec/parser/options_spec.rb
```diff
@@ -257,6 +257,7 @@ def remaining
         expect(parse("-f=12")["foo"]).to eq("12")
         expect(parse("--foo=12")["foo"]).to eq("12")
         expect(parse("--foo=bar=baz")["foo"]).to eq("bar=baz")
+        expect(parse("--foo=-bar")["foo"]).to eq("-bar")
```
rafaelfranca: We need a test when there is more than one options.

lib/thor/parser/options.rb
```diff
@@ -83,7 +84,11 @@ def parse(args) # rubocop:disable MethodLength
             when SHORT_SQ_RE
               unshift($1.split("").map { |f| "-#{f}" })
               next
-            when EQ_RE, SHORT_NUM
+            when EQ_RE
+              unshift($2)
+              switch = $1
+              @force_value = true
```
rafaelfranca: I think setting this state here will broke it if there are more options after the first with `-` in the value, can you add tests to make sure it will work?

PR #500, opened 2015-09-28
fix 'thor_script subcommand [args] --help' 🌈 

doudou:
Currently, doing so leads to

    [~/dev/flat_fish]% autoproj test --help
    Usage:
      autoproj help [COMMAND]

instead of the expected result of "autoproj help test"

This case is handled internally by thor by transforming the subcommand invocation into `help *ARGV`

There were two problems:
 - first invoke was called with (klass, args, options, parameters) while invoke expects (klass, subcommand, args, options, parameters)
 - the --help argument was kept, which ends up as `help *ARGV --help, i.e. "give me the help of the 'help' subcommand

The commit removes the --help option before calling the subcommand, and fixes the call signature. It only does so for the "massaged" 'help' subcommand because invoke() has a way-too-weird call semantic for me to change the call in the general case without fearing to break something else.

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -235,8 +235,11 @@ class Thor # rubocop:disable ClassLength
 
       define_method(subcommand) do |*args|
         args, opts = Thor::Arguments.split(args)
-        args.unshift("help") if opts.include? "--help" or opts.include? "-h"
-        invoke subcommand_class, args, opts, :invoked_via_subcommand => true, :class_options => options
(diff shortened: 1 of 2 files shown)


doudou:
Is 1.8.7 really still a target platform ????

sferik:
@doudou It is.

doudou:
OK then ... will modify the PR

doudou:
@sferik: fixed for 1.8.7

PR #503, opened 2015-10-06
fix incorrect command name

miya0001:
Incorrect command name on `handle_argument_error`.

For example:

```
$ foo bar baz "hello"
ERROR: "foo baz" was called with arguments ["hello"]
Usage: "foo bar baz"
```


rafaelfranca:
Could you add a test case?

rafaelfranca:
Closing since there was no update.

PR #505, opened 2015-10-24
Feature/dont replace nonexistant files

vcavallo:
🌈
^ the rainbow of compliance :)

I came across some difficulty earlier today as a rails generator was attempting to inject into a file that didn't exist in my project. The error was relatively clear but I was surprised to find that Thor didn't handle this issue. My pull request helps that a little, hopefully, and would address at least [this guy's](https://github.com/erikhuda/thor/issues/463) issue.
Please let me know if the specs are inadequate or if this needs documentation.

### DIFF
--- a/lib/thor/actions/inject_into_file.rb
+++ b/lib/thor/actions/inject_into_file.rb
@@ -57,7 +57,7 @@ class Thor
           replacement + '\0'
         end
 
-        if exists?
+        if File.exist?(destination)
           replace!(/#{flag}/, content, config[:force])
         else
           raise MalformattedArgumentError, "The file #{ destination } does not appear to exist"
--- a/lib/thor/actions/inject_into_file.rb
+++ b/lib/thor/actions/inject_into_file.rb
@@ -57,7 +57,7 @@ class Thor
           replacement + '\0'
         end
 
-        if File.exist?(destination)
+        if exists?
           replace!(/#{flag}/, content, config[:force])
         else
           raise MalformattedArgumentError, "The file #{ destination } does not appear to exist"
--- a/lib/thor/actions/inject_into_file.rb
+++ b/lib/thor/actions/inject_into_file.rb
@@ -1,4 +1,5 @@
 require "thor/actions/empty_directory"
+require "pry"
 
 class Thor
   module Actions
@@ -56,7 +57,11 @@ class Thor
           replacement + '\0'
         end
 
-        replace!(/#{flag}/, content, config[:force])
+        if File.exist?(destination)
+          replace!(/#{flag}/, content, config[:force])
+        else
+          raise MalformattedArgumentError, "The file #{ destination } does not appear to exist"
+        end
       end
 
       def revoke!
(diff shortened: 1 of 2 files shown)


vcavallo:
Specs pass for me. If someone wants to give me a tip on why CI is failing I'll happily address.

xentek:
LGTM :+1:

vcavallo:
Thanks! I was just wondering about this pull request a couple days ago

lib/thor/actions/inject_into_file.rb
```diff
@@ -56,7 +56,11 @@ def invoke!
           replacement + '\0'
         end
 
-        replace!(/#{flag}/, content, config[:force])
+        if File.exist?(destination)
+          replace!(/#{flag}/, content, config[:force])
+        else
+          raise MalformattedArgumentError, "The file #{ destination } does not appear to exist"
```
rafaelfranca: This exception is from when the argument can't be parsed. This should raise `Thor::Error`.

spec/actions/inject_into_file_spec.rb
```diff
@@ -70,6 +70,21 @@ def file
       expect(File.read(file)).to eq("__start__\nREADME\nmore content\n__end__\n")
     end
 
+    it "does not attempt to change the file if it doesn't exist" do
+      expect_any_instance_of(Thor::Actions::InjectIntoFile).not_to receive(:replace!)
```
rafaelfranca: I think we should assert that the file was not created.

spec/actions/inject_into_file_spec.rb
```diff
@@ -70,6 +70,21 @@ def file
...
+      invoker.inject_into_file "idontexist", :before => "something" do
+        "any content"
+      end rescue nil
+    end
+
+    it "raises a malformatted argument error including filename if file doesn't exist" do
+      expect do
+        invoker.inject_into_file "idontexist", :before => "something" do
```
rafaelfranca: Can we use `invoke!` here?

vcavallo:
Hi @rafaelfranca! I've addressed your comments in the recent commit. Thanks for taking the time to take a look at this. 👍

spec/actions/inject_into_file_spec.rb
```diff
@@ -70,6 +70,22 @@ def file
       expect(File.read(file)).to eq("__start__\nREADME\nmore content\n__end__\n")
     end
 
+    it "does not attempt to change the file if it doesn't exist" do
+      invoker.inject_into_file "idontexist", :before => "something" do
+        "any content"
+      end rescue nil
```
rafaelfranca: Maybe instead of rescue here we can merge both tests and assert that the file was not created in the test below?

vcavallo: I thought it was better to have separate `expect`s for the different aspects of the change, but I've adjusted the test file according to your comments now. let me know if you had something different in mind.
Thanks.

vcavallo:
Let me know if you'd like to see any other changes @rafaelfranca

PR #508, opened 2015-11-15
ran rubocop autocorrect

Bobby-H:


rafaelfranca:
Thanks! We already did it.

PR #509, opened 2015-12-12
replaced '.map{|_,o| o}' on hash with '.values'

humzashah:
The values of a hash someplace were being obtained using a non-standard method, as if the hash was a 2D array. This has been replaced with the `.values` method now. Tests have been checked locally on Ruby version 2.2.3. They're passing. Travis will execute them on remaining environments.


sferik:
Thanks!

PR #510, opened 2015-12-12
specify addressable version on ruby 1.8.7 platform

humzashah:
webmock required addressable version 2.3.6 or greater. a later
has now required minimum ruby version to be 1.9. this commit
specified which addressable version to use to travis tests pass
while using ruby 1.8.7


humzashah:
rebasing feature on this commit, so closing PR.

PR #512, opened 2015-12-19
Force ERB in Thor::Actions to use String buffer

k0kubun:
`#capture` depends on ERB's internal buffer. Since ERB's default buffer is changed to Array buffer in https://github.com/ruby/ruby/pull/1143, I fixed a buffer of Thor::Actions' erb to String buffer.

### DIFF
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -113,7 +113,7 @@ class Thor
       context = config.delete(:context) || instance_eval("binding")
 
       create_file destination, nil, config do
-        content = ERB.new(::File.binread(source), nil, "-", "@output_buffer").result(context)
+        content = CapturableERB.new(::File.binread(source), nil, "-", "@output_buffer").result(context)
         content = block.call(content) if block
         content
       end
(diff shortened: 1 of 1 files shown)


yui-knk:
see [issue on Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/11842)

yui-knk:
I think it is better to split first commit to another PR. The problem first commit will fix and the problem second commit will fix are differ.

k0kubun:
> I think it is better to split first commit to another PR. The problem first commit will fix and the problem second commit will fix are differ.

I think so too. Done https://github.com/erikhuda/thor/pull/513.

yui-knk:
:heart_eyes:

sferik:
Thanks for making this change in advance of the Ruby 2.3 release. It looks like the build is passing without merging https://github.com/erikhuda/thor/pull/513. Can that be closed?

k0kubun:
Thank you for merging this!

> It looks like the build is passing without merging #513. Can that be closed?

Not sure. While builds jobs are finished in https://travis-ci.org/erikhuda/thor/builds/97829708, a build for ruby-head (Ruby 2.3) haven't been finished yet https://travis-ci.org/erikhuda/thor/jobs/97829718. 

I guess it needs to update a test or implementation to finish all tests in Ruby 2.3.

PR #513, opened 2015-12-19
Use block instead of to_proc

k0kubun:
Test suite stops at `$stdin.noecho(&:gets)` with Ruby 2.3 (ruby-head build on travis CI).
When I changed it to `$stdin.noecho { |stdin| stdin.gets }`, it worked somehow.
But I don't know why... :(

### DIFF
--- a/lib/thor/line_editor/basic.rb
+++ b/lib/thor/line_editor/basic.rb
@@ -23,7 +23,9 @@ class Thor
         if echo?
           $stdin.gets
         else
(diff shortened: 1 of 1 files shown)


sferik:
This seems like a bug in Ruby 2.3. Not an issue specific to this library.

k0kubun:
> This seems like a bug in Ruby 2.3. Not an issue specific to this library.

I guess so too. So closing for now.

PR #521, opened 2016-08-15
Change default exit code behavior to match norm

### LINKED ISSUE #244: Returns exit 0 when a required argument is not provided
Simple example

```
#!/usr/bin/env ruby
require 'thor'

class Foo < Thor

  desc "foo", ""
  method_option :output, required: true
  def foo
    puts options[:output]
  end

end

Foo.start
```

```
$ ./foo.rb foo -output=blah                                                                                                                               [1s] 
No value provided for required options '--output'
$ echo $?                                                                                                                                                 [0s] 
0
```

I'd expect this to return an error exit code

dkniffin:
Fixes #244

### DIFF
--- a/lib/thor/base.rb
+++ b/lib/thor/base.rb
@@ -621,7 +621,7 @@ class Thor
 
       # A flag that makes the process exit with status 1 if any error happens.
       def exit_on_failure?
-        false
+        true
       end
 
       #


rafaelfranca:
As you can see it breaks the thor test suite. We need to make sure everything passes.

teaforthecat:
Dang, it looks like this is failing due to a 500 response from coveralls.io . All the tests actually pass. I wonder if the run could be retried, or maybe this branch could be rebased, which would retry the run.

rafaelfranca:
No, it is not failing because of coveralls.io. There is a legit test failing:

```
1) Thor::Group#invoke_from_option with default type allows to invoke a class from the class binding by the given option
     Failure/Error: exit(1) if exit_on_failure?
     
     SystemExit:
       exit
     # ./spec/../lib/thor/base.rb:443:in `exit'
     # ./spec/../lib/thor/base.rb:443:in `start'
     # ./spec/group_spec.rb:132
     # ./spec/helper.rb:59:in `capture'
     # ./spec/group_spec.rb:131
     # ./spec/helper.rb:59:in `capture'
     # ./spec/group_spec.rb:130
```

teaforthecat:
Oh, sorry, it is just on the 1.8.7 build, I see.

dkniffin:
I'm sorry I made this PR and just abandoned it. I'm going to work on fixing these tests now.

dkniffin:
Alright, we've got a green build. @rafaelfranca can you take another look at this?

Also, I just noticed #578. I'm not sure if/how this relates to that change. Maybe @jmax315 can elaborate

jmax315:
It looks like, on casual inspection, we both fixed the same two problems, although we came at it from two different symptoms. It may still be worth pulling in my specs though (spec/fixtures/exit_status.thor and spec/script_exit_status.rb), as they directly address the symptom that started me off.
Merged the current main head into my fork, and my specs pass, but two unrelated specs fail (both in spec/actions/file_manipulation_spec.rb; lines 12 and 86). I'll look into what's going on here later and get back to you.

PR #528, opened 2016-10-10
Fix injecting into file which contains chinese

lingceng:
🌈 Also work for other utf-8 codes such as Japanese, Korean.


pinglamb:
👍

PR #531, opened 2016-11-02
Make HashWithIndifferentAccess fail gracefully...

jfinkhaeuser:
... when trying to access a key as a method, but the key does not exist.

Before this PR, it would simply return nil, i.e.

```ruby
x = HashWithIndifferentAccess.new
x.does_not_exist # would have succeeded without error or warning
```

After this PR:

```ruby
x = HashWithIndifferentAccess.new "foo" => "bar"
x.foo # => "bar"
x.does_not_exist # raises NoMethodError
```

### DIFF
--- a/lib/thor/actions.rb
+++ b/lib/thor/actions.rb
@@ -305,14 +305,18 @@ class Thor
       super.merge!(:destination_root => destination_root)
     end
 
+    FORCE_SKIP_OPPOSITES ||= { force: "skip", skip: "force" }.freeze
+
     def _cleanup_options_and_set(options, key) #:nodoc:
       case options
       when Array
         %w[--force -f --skip -s].each { |i| options.delete(i) }
         options << "--#{key}"
+        options << "--no-#{FORCE_SKIP_OPPOSITES[key.to_sym]}"
       when Hash
         [:force, :skip, "force", "skip"].each { |i| options.delete(i) }
         options.merge!(key => true)
+        options.merge!(FORCE_SKIP_OPPOSITES[key.to_sym] => false)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


rafaelfranca:
Tests are broken after this change so it seems we are expecting it to behave like that. Could you investigate?

jfinkhaeuser:
I ran the tests and they worked... 
On 5 Nov 2016 6:51 pm, Rafael França <notifications@github.com> wrote:Tests are broken after this change so it seems we are expecting it to behave like that. Could you investigate?

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  






!DSPAM:1,581e1c6e23229948712254!

jfinkhaeuser:
I will investigate tomorrow

jfinkhaeuser:
I don't know what happened there, I could swear I ran the tests and nothing broke. Huh. Obviously that wasn't the case.

Anyhow... yes, you were relying on it. The change I made to Actions isn't exactly sexy, but IMHO it's still a good one.

My point of view is that something as generic as HashWithIndifferentAccess should behave sanely for its purpose, and if Actions wants to override some behaviour, it has to do so explicitly. YMMV.

jfinkhaeuser:
The remaining failures are to do with `term-ansicolor` not installing.

piotrmurach:
I have submitted PR to fix the issue with `coveralls` https://github.com/lemurheavy/coveralls-ruby/pull/108

rafaelfranca:
After further consideration I believe the actual behavior make sense. It behaves in the same way as `[]` to a unknown key.

Thank you so much for the pull request.

PR #537, opened 2016-12-17
Remove unused argument in the method Thor#banner

mgrachev:


rafaelfranca:
Thank you for the pull request but `banner` is part of the public API and we can't change the signature of the method like this. You can also find a implementation of banner with three argument in Thor::Runner.

PR #539, opened 2017-01-15
Lazy-load io/console to avoid activating the gem unecessarily

segiddins:


segiddins:
Ping on getting this reviewed? Thanks!

piotrmurach:
Hey, have you found it to be performance issue? I'm curious to see whether hiding file dependency in the bowels of codebase is good enough tradeoff?

rafaelfranca:
The main reason to lazy loading it is that it is being extracted to a gem in Ruby 2.4+ and it may not be present in a ruby installation. It also is only necessary if the ruby installation doesn't have readline.

This change does make sense. Thanks.

segiddins:
Its not a performance issue, but rather an issue for bundler & other gems -- if io-console is eagerly loaded, then the user cannot control what version is used

piotrmurach:
I'm asking cuz I'm in a similar boat relying on 'io-console' and in particular in changing echo mode. I guess I'm looking for a robust solution. For example, what will happen if you're running on Ruby 2.4 and `io-console` gem is not installed whether through bundler or rubygems? Will the application explode with `LoadError` when that particular branch is called?

rafaelfranca:
Yeah, that is what would happen today. We can improve it though with a fallback.

PR #541, opened 2017-01-25
Fix "file clash" bug

nodo:
## Summary

In some cases creating a file would raise an exception, due to clashes with existing file. E.g. creating a file where a directory with the same name is present.

## How to reproduce the bug

```
$ mkdir err
$ touch er/hello
$ irb
>> require 'thor'; class A < Thor; include Thor::Actions; end; a = A.new; a.create_file('./err/hello/boom')
      create  err/hello/boom
Errno::EEXIST: File exists @ dir_s_mkdir - /Users/anodari/other_projects/thor/err/hello
from /Users/anodari/.rbenv/versions/2.4.0/lib/ruby/2.4.0/fileutils.rb:228:in `mkdir'
```

## What happen using the bug fix
```
>> require './lib/thor'; class A < Thor; include Thor::Actions; end; a = A.new; a.create_file('./err/hello/boom')
  file_clash  err/hello/boom
```

I would be more than happy to discuss further this issue. Feel free to comment about the naming and the approach.

### DIFF
--- a/lib/thor/actions/empty_directory.rb
+++ b/lib/thor/actions/empty_directory.rb
@@ -112,11 +112,17 @@ class Thor
         if exists?
           on_conflict_behavior(&block)
         else
-          say_status :create, :green
           yield unless pretend?
+          say_status :create, :green
         end
 
         destination
+      rescue Errno::EISDIR, Errno::EEXIST
+        on_file_clash_behavior
+      end
+
+      def on_file_clash_behavior
+        say_status :file_clash, :red
       end
 
       # What to do when the destination file already exists.
--- a/spec/actions/create_file_spec.rb
+++ b/spec/actions/create_file_spec.rb
@@ -131,6 +131,30 @@ describe Thor::Actions::CreateFile do
         end
       end
     end
+
+    context "when file exists and it causes a file clash" do
+      before do
+        create_file("doc/config")
+        invoke!
+      end
+
+      it "generates a file clash" do
+        create_file("doc/config/config.rb")
+        expect(invoke!).to eq("  file_clash  doc/config/config.rb\n")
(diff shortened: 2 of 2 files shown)


nodo:
Regarding the travis checks:
- The failure for ruby 2.3.3 seems to be a known issue of `rainbow` gem: https://github.com/sickill/rainbow/issues/49

lib/thor/actions/empty_directory.rb
```diff
@@ -112,11 +112,18 @@ def invoke_with_conflict_check(&block)
         if exists?
           on_conflict_behavior(&block)
         else
-          say_status :create, :green
```
rafaelfranca: Why this was changed?

lib/thor/actions/empty_directory.rb
```diff
@@ -112,11 +112,18 @@ def invoke_with_conflict_check(&block)
...
           yield unless pretend?
+          say_status :create, :green
         end
 
         destination
+      rescue SystemCallError => e
```
rafaelfranca: It is better to rescue the specific exception here and avoid the `unless` instead of the parent class.

nodo:
Thanks for your review @rafaelfranca, sorry for the delay. I will address your comments in the next days.

lib/thor/actions/empty_directory.rb
nodo: If `say_status` is before `yield`, in case of a conflict we have a misleading message. For instance:
```
create  doc/config/config.rb
file_clash  doc/config/config.rb
```

nodo:
@rafaelfranca Thanks again for your review. I have addressed you comments and rebase my branch.

PR #542, opened 2017-01-27
Bug Fix on Windows changing /USER~1/ to /User Name/

garrrth:
🌈

This is to fix a bug that I (and others, see [rspec#1555](https://github.com/rspec/rspec-rails/issues/1555), [rspec#1577](https://github.com/rspec/rspec-rails/issues/1577), [rspec#1747](https://github.com/rspec/rspec-rails/issues/1747) and [rails#27129](https://github.com/rails/rails/issues/27129)) have experienced when running `rails generate rspec:install` on a windows machine.

Ruby creates a tmpdir with `/USER~1/` and then thor uses another method, `files(lookup)`, that converts it to `/User Name/` and tries to compare the two and gsub it out when it assigns it to `file_destination`. My fix, though a little hacky (I'm new to Ruby and don't really know of a better way in all honesty) works and shouldn't create problems elsewhere as far as I can see.

### DIFF
--- a/lib/thor/actions/directory.rb
+++ b/lib/thor/actions/directory.rb
@@ -80,6 +80,11 @@ class Thor
         files(lookup).sort.each do |file_source|
           next if File.directory?(file_source)
           next if config[:exclude_pattern] && file_source.match(config[:exclude_pattern])
+
+          # Fix for Windows, turns /USER~1/ into /User Name/ for temp directory so that
+          # gsub will find a match
+          source.gsub!(/\/\w*~1\//, "/"+ENV['USERNAME']+"/")
+
           file_destination = File.join(given_destination, file_source.gsub(source, "."))
           file_destination.gsub!("/./", "/")
 


piotrmurach:
Hey, great that you have looked into this. I'm a bit concerned about the fix because the regex assumes `~1` to proceed username but what about legitimate file paths that contain such characters? Also I would encourage you to write a test that demonstrates this problem/fix.

garrrth:
Hey Piotr, thanks for your feedback. I have a different approach to this now - replacing the file path that is created by `Dir.glob` with it's original stored in `lookup`. I also have it only performing the replacement when it sees a windows environment variable for username (which I can imagine is the only use case for this). I'll write a test before pushing it up, but that will be after work 😞 !

garrrth:
I'm questioning myself right now as to whether this is the right fix!! It is essentially a work around rather than a fix. For example, this issue could be remedied in this part on thor, or on rspec which is creating the tmpdir file path before it is sent to thor.

It is essentially `Dir.glob(lookup, File::FNM_DOTMATCH)` that is making the change from `USER~1` to `User Name`, so is there another method that could replace this that doesn't make that change??

garrrth:
Ok, digging more unearths that this is something that has arisen since Ruby 2.3. See Ruby Issue [#10819](https://bugs.ruby-lang.org/issues/10819) referenced from Puppet pull-request [#5156](https://github.com/puppetlabs/puppet/pull/5156).

Also struggling to figure out what exactly I need to test for with this (I'm new to writing tests and all that!).

TraGicCode:
Any uodate in this

mathieurousseau:
Any update on this? still can't run rspec:install on windows machine

MarcoDeJesus:
Followed the work around and updated the file by my own and I'm able to continue working. Thanks @garrrth

Issue #363, opened 2013-09-03
required class option makes help command less useful


b-dean:
When a `Thor` class has required class options, it makes the `help` command also require them. This means that the user can't run `cli help COMMAND` to get help on that command because the required options are not provided. Here's an example:

```ruby
require 'thor'
class Example < Thor
  class_option :username, :desc => 'Some username', :required => true
  class_option :password, :desc => 'Some password', :required => true

  desc 'do_something', 'Do something'
  def do_something
    ExternalSystem.new(options[:username], options[:password]).do_something
  end

  desc 'something_else', 'Some other thing'
  def something_else
    ExternalSystem.new(options[:username], options[:password]).something_else
  end
end

Example.start(%w{help do_something})
```

I would expect that to give some help for the `do_something` command so the user knows what options are needed. Instead I just get:

    No value provided for required options '--username', '--password'

Is there some way to make these class options not apply to the `help` command?

kaelumania:
whats the status of this issue?

b-dean:
I don't know. As you can see #462 has been sitting there for a while.

rspencer01:
Bug is still biting.  Please fix!

sferik:
I just merged #462. I’ll see if I can put a release together soon.

sferik:
In the mean time, can everyone please verify that this problem is fixed in `master`?

nishidy:
I tried the master and it just worked with class_option and required:true together but failed on 0.19.1 with Thor::RequiredArgumentMissingError.

xam7247:
Does this work for anyone ? I am still facing this issue, even when I make method_options required. Thor v0.19.1

omgreenfield:
'Twould be great if, when we close a ticket like this, we provide an example of it working instead of hoping people can figure out by inspecting the code. I can't get this feature to work

PR #545, opened 2017-02-13
Add disable_required_check! and make disable the check in help

rafaelfranca:
In Thor 0.19.2 it was added a new class method `disable_class_options`.
It was added to disable the check for required options in the help command. This new method disabled the option parser for help, so it is not possible to pass option to that command anymore.

That change introduced a regression and it is fixing the problem at the wrong level. What we want is to disable the required check not the option parsing.

So, we reverted the `disable_class_options` change and added a new method `disable_required_check!` to disable the required check in the commands we want.

Theoretically this is a breaking change but the `disable_class_options` introduction was also a breaking change so it is better to revert it and fix the problem at the correct place.

Related with #462 and #363.

See https://github.com/bundler/bundler/pull/5221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## 0.20.0
+* Add `disable_required_check!` to disable check for required options in some commands.
+  It is a substitute of `disable_class_options` that was not working as intended.
+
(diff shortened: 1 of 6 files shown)


sferik:
👍

PR #551, opened 2017-03-07
Fix double negatives on bool opts with underscores

tommarshall:
#### Because:

* #460 (released in `0.19.2`) introduced a change that prevents double negative options (`--no-no-foo`) for already negative  boolean options (`--no-foo`).
* That change checks if the option name string starts with 'no-', before it's `dasherized`, therefore negative options defined with underscores (`:no_foo`) still generate a double negative `--no-no-foo` option.

#### This change:

* Adds a test for, and fixes the double negative boolean options
  defined with underscores.

### DIFF
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -88,7 +88,7 @@ class Thor
       sample = "[#{sample}]" unless required?
 
       if boolean?
-        sample << ", [#{dasherize('no-' + human_name)}]" unless (name == "force") || name.start_with?("no-")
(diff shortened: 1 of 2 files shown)


piotrmurach:
This is related to https://github.com/erikhuda/thor/issues/553 in as far as negation of boolean flags is concerned. I feel as though this automatic negation provides more obscure behaviour than real benefits. Your fix kind of demonstrates the possible corner cases etc.... I'm not sure if this should be merged or the behaviour of actual negation changed. Teetering towards the later.

PR #558, opened 2017-04-02
use latest ruby version in CI

takkanm:


.travis.yml
```diff
@@ -9,8 +9,8 @@ rvm:
   - 2.0.0
   - 2.1.10
   - 2.2.6
```
koic: Ruby 2.2.7 has been released and available on Travis CI. Is there reason to not upgrade this version?
https://www.ruby-lang.org/en/news/2017/03/28/ruby-2-2-7-released

rafaelfranca: @takkanm could you add the 2.2.7 too?

takkanm: ok, I'll update it.

PR #561, opened 2017-05-04
Updating rake and rdoc to recent versions

blazeeboy:
Gemnasium shows that runtime dependencies needed updates for both rake and rdoc,
updated to the latest versions, tests are green, I hope I'm not missing anything.


blazeeboy:
aha, rake can't be updated as it doesn't support older ruby, well time to close that PR then 😄

PR #563, opened 2017-06-06
Prevent infinite loop on file collision when stdin closed

tpope:
🌈

If stdin is closed, any attempt to read user input returns `nil`, resulting in an infinite loop on file collision since that case isn't accounted for. The solution I have submitted treats `nil` the same as pressing enter. I'm also outputting a newline to prevent subsequent output from ending up on the same line.


rafaelfranca:
Thanks

PR #565, opened 2017-06-19
Avoid extra spaces from nil/false switches in Options.to_switches

segiddins:


segiddins:
I don't know why this is only decreasing coverage on the one ruby version, given I added tests for the lines I touched

rafaelfranca:
Thanks!

PR #566, opened 2017-06-20
refactor ordered_hash_spec

colby-swandale:
This PR refactors the `ordered_hash_spec` to keep with convention of the rest of the suite and added a few new specs.

There was a few issues i felt need addressing in this spec.
* wrap the described behaviour in `describe` blocks with the method name that is being tested.
* Don't use an instance variable as the test subject and don't depend on the state of the subject from previous specs

Thanks


rafaelfranca:
Awesome refactoring

PR #567, opened 2017-07-02
Updates so things work with frozen string literals

pat:
These changes get Thor working when frozen string literals are enabled.

Regarding test suite dependencies and their compatibility with frozen string literals:

* RSpec is compatible in their latest git commits (for `rspec-core`, `rspec-expectations`, etc).
* `simplecov` and its dependency `simplecov-html` are compatible in their latest git commits.
* Addressable is compatible in this PR: https://github.com/sporkmonger/addressable/pull/260
* Webmock is not yet compatible - I'm working on it, though I've also uncovered a frozen string literal bug in stdlib. The patches I've made locally (not yet submitted to a PR) are enough to get Thor's build green though.

### DIFF
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -341,7 +341,7 @@ class Thor
     end
 
     def with_output_buffer(buf = "".dup) #:nodoc:
-      raise "No: #{buf}" if buf.frozen?
+      raise ArgumentError, "Buffer can not be a frozen object" if buf.frozen?
       old_buffer = output_buffer
       self.output_buffer = buf
       yield
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -340,7 +340,8 @@ class Thor
       with_output_buffer { yield(*args) }
     end
 
-    def with_output_buffer(buf = "") #:nodoc:
+    def with_output_buffer(buf = "".dup) #:nodoc:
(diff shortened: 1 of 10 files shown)


pat:
With regards to the stdlib bug (net/protocol) - looks like it's fixed in trunk, though has not (yet) been backported into the next 2.4.x release.

I'm happy to add the pragma comment (in this PR or a new one) to at least ensure regressions aren't introduced, if you'd like?

segiddins:
Maybe just add the magic comment to the Thor source files?

pat:
@segiddins: I'm certainly happy to make that change if it's the maintainers' preference, just didn't wish to presume, because it involves editing pretty much every file, and then needs to be remembered for any files that get added to the project further down the track.

lib/thor/actions/file_manipulation.rb
```diff
@@ -340,7 +340,8 @@ def capture(*args)
       with_output_buffer { yield(*args) }
     end
 
-    def with_output_buffer(buf = "") #:nodoc:
+    def with_output_buffer(buf = "".dup) #:nodoc:
+      raise "No: #{buf}" if buf.frozen?
```
rafaelfranca: Maybe a better descriptive error here. Also I think the exception should be an ArgumentError

pat: Thanks @rafaelfranca - that was actually a debugging line that I shouldn't have committed in the first place. Have kept the raise, using ArgumentError, and put in a more useful error message.

PR #569, opened 2017-07-06
Add `check_default_type!`

rafaelfranca:
It check if the default value of an option matches the defined type.

This removes the warning on usage since there is no much what the command users could do about it and give the command authors the possibility to check for programming errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## 0.20.0
+* Add `check_default_type!` to check if the default value of an option matches the defined type.
+  It removes the warning on usage and gives the command authors the possibility to check for programming errors.
+
 * Add `disable_required_check!` to disable check for required options in some commands.
   It is a substitute of `disable_class_options` that was not working as intended.
 
--- a/lib/thor/base.rb
+++ b/lib/thor/base.rb
@@ -151,6 +151,21 @@ class Thor
         !!check_unknown_options
       end
 
+      # If you want to raise an error when the default value of an option does not match
+      # the type call check_default_type!
+      # This is disabled by default for compatibility.
+      def check_default_type!
+        @check_default_type = true
+      end
+
+      def check_default_type #:nodoc:
+        @check_default_type ||= from_superclass(:check_default_type, false)
+      end
+
+      def check_default_type? #:nodoc:
+        !!check_default_type
+      end
+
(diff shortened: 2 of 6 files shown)


lib/thor/base.rb
```diff
@@ -151,6 +151,21 @@ def check_unknown_options?(config) #:nodoc:
         !!check_unknown_options
       end
 
+      # If you want to raise an error when the default value of an option does not match
+      # the type call check_default_type!
+      # This is disabled by default.
```
matthewd: \+ "for compatibility"

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## 0.20.0
+* Add `check_default_type!` to check if the deafault value of an option matches the defined type.
+  It removes the warning on usage and give the command authors the posibility to check for programing erros.
```
matthewd: Few typos in here: default, possibility, errors

lib/thor/parser/option.rb
```diff
@@ -4,13 +4,14 @@ class Option < Argument #:nodoc:
 
     VALID_TYPES = [:boolean, :numeric, :hash, :array, :string]
 
-    def initialize(name, options = {})
+    def initialize(name, options = {}, check_default_type = false)
```
matthewd: Should this be passed in `options`?

spec/thor_spec.rb
```diff
@@ -583,6 +583,24 @@ def unknown(*args)
       expect(klass.start(%w(unknown foo --bar baz))).to eq(%w(foo))
     end
 
+    it "does not checks the default type" do
```
andremedeiros:
```ruby
it "does not check the default type when check_default_type! is not called" do
```

lib/thor/parser/option.rb
andremedeiros: Sounds like `options` is where it belongs, yeah.

rafaelfranca:
Fixed all the points

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## 0.20.0
+* Add `check_default_type!` to check if the default value of an option matches the defined type.
+  It removes the warning on usage and give the command authors the possibility to check for programming errors.
```
andremedeiros: ... and `gives` the command ...

PR #572, opened 2017-07-28
Some minor code improvements

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/lib/thor/shell/basic.rb
+++ b/lib/thor/shell/basic.rb
@@ -1,6 +1,8 @@
 class Thor
   module Shell
     class Basic
+      DEFAULT_TERMINAL_WIDTH = 80
+
       attr_accessor :base
       attr_reader   :padding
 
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request. We don't return singleton `true`/`false` in methods ending with `?` since in Ruby anything that is not `nil` or `false` is true.

lib/thor/shell/basic.rb
```diff
@@ -276,14 +276,15 @@ def file_collision(destination)
       # This code was copied from Rake, available under MIT-LICENSE
       # Copyright (c) 2003, 2004 Jim Weirich
       def terminal_width
+        DEFAULT_WIDTH = 80
```
rafaelfranca: To this constant make sense it can't be inside the method.

sschuberth:
> We don't return singleton `true`/`false` in methods ending with `?` since in Ruby anything that is not `nil` or `false` is true.

While this is syntactically correct, this is considered bad practice, see e.g. https://github.com/bbatsov/ruby-style-guide#bool-methods-qmark. Also, IIRC I found this because it was flagged by some static code analyzer (I don't remember which one).

lib/thor/shell/basic.rb
sschuberth: Why, if it's not used elsewhere (yet)?

rafaelfranca:
> While this is syntactically correct, this is considered bad practice

What is bad practice or not is defined by the project maintainer. We don't follow the bbatsov's ruby-stile-guide in that rule, neither Rails does, so could you remove that commit?

sschuberth:
> What is bad practice or not is defined by the project maintainer.

Yeah, in a world of alternative facts that's probably true :wink:

> We don't follow the bbatsov's ruby-stile-guide in that rule, neither Rails does, so could you remove that commit?

Done, anyway.

rafaelfranca:
Thank you so much for doing those kinds of improvements. Really appreciated

PR #578, opened 2017-09-19
Fix incorrect use of Process::exit. This fixes open issue #244.

jmax315:
🌈
Unlike the C library function exit(), Ruby's version takes a boolean,
not an integer. Added a spec to check the actual shell-level exit
value of a Thor script, and fixed calls to Process::exit.

### DIFF
--- a/lib/thor/base.rb
+++ b/lib/thor/base.rb
@@ -466,13 +466,13 @@ class Thor
         dispatch(nil, given_args.dup, nil, config)
       rescue Thor::Error => e
         config[:debug] || ENV["THOR_DEBUG"] == "1" ? (raise e) : config[:shell].error(e.message)
-        exit(1) if exit_on_failure?
+        exit(false) if exit_on_failure?
       rescue Errno::EPIPE
         # This happens if a thor command is piped to something like `head`,
         # which closes the pipe when it's done reading. This will also
         # mean that if the pipe is closed, further unnecessary
         # computation will not occur.
-        exit(0)
+        exit(true)
       end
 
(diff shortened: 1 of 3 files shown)


jmax315:
The build for 1.8.7 fails with problems in the webmock gem. As far as I can see, my change didn't cause these; does the current head build for 1.8.7 in CI correctly?

lostapathy:
#583 fixes the travis error that shows up here.

deivid-rodriguez:
@rafaelfranca What do you think about this? Backwards compatibility considerations about changing the default for this setting apart, this is fixing an actual bug, right?

deivid-rodriguez:
Would it help if I rebase this PR and get it green? According to https://github.com/erikhuda/thor/issues/244#issuecomment-278358052, even setting `exit_on_failure?` to `true` does not exit on failure. This PR seems like a feasible explanation and fix for that issue, and it's backwards compatible.

deivid-rodriguez:
@rafaelfranca Since you're having a look at `thor` I'd like to politely ping you about this again.

deivid-rodriguez:
Thanks! :heart:

The tests need to be skipped on 1.8.7 due to using `spawn` (or find another version that works ok on 1.8.7). Skipping seems easiest, no?

deivid-rodriguez:
Funny, while creating the PR to skip the tests on 1.8.7 I tried reverting the code changes in this PR and the specs still passed. So it seems that this PR didn't fix anything and the `exit_on_failure?` thing is actually working fine? The tests still seem good to catch future regressions so I guess it's ok.

PR #582, opened 2017-10-01
Allow user to ignore class options for command

snehitgajjar:
There are two scenario where this funcitonality can be useful.

1. There is a `class_option` with `required: true` and some command does not need that class option.
2. There are some class options(`required: true`) which is used by most of the commands but not necessary for other commands.


snehitgajjar:
The build for 1.8.7 fails with problems in the webmock gem. As far as I can see, my change didn't cause these. It is erring out on requiring `webmock/rspec` in this file `./spec/actions/create_file_spec.rb`.

lostapathy:
#583 fixes the webmock dep issue that's causing travis to fail here.

snehitgajjar:
@rafaelfranca Can you please take a look at it?

PR #584, opened 2017-10-07
update ruby version in travis config

lostapathy:


lostapathy:
The build failure on this gets fixed by #583.

PR #588, opened 2017-11-29
Remove the root path from the absolute path only once

CvX:
🌈In cases where root path is also a substring of a relative path, it was being erroneously removed twice (or more).

e.g. for absolute path `/app/config/application.rb` when root is `/app` it would display the relative path as `config.lication.rb` instead of `config/application.rb`.


MaxLap:
I made a PR #620 which fixes this and other edge cases with this method. The original commit of this PR is kept intact in my PR.

MaxLap:
#620 was merged. This can be closed

PR #589, opened 2017-12-03
Use correct line numbers for `class_eval` and `module_eval` methods

pocke:
🌈 When using heredoc, the line number that is passed to `eval` method should be a `__LINE__ + 1`.

For example

```ruby
instance_eval <<-END, __FILE__, __LINE__
  raise 'In the instance_eval'
END
```

It print below.

```bash
$ ruby test.rb
test.rb:1:in `<main>': In the instance_eval (RuntimeError)
	from test.rb:1:in `instance_eval'
	from test.rb:1:in `<main>'
```

It says "This exception occurred on the 1st line", but the `raise` is on the 2nd line.

This change will fix this problem.

### DIFF
--- a/lib/thor/group.rb
+++ b/lib/thor/group.rb
@@ -61,7 +61,7 @@ class Thor::Group
         invocations[name] = false
         invocation_blocks[name] = block if block_given?
 
-        class_eval <<-METHOD, __FILE__, __LINE__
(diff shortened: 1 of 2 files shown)


marcandre:
Looks like a quick merge.

marcandre:
:+1: @pocke & @rafaelfranca 😸

PR #594, opened 2018-03-18
Deprecate safe_level of ERB.new in Ruby 2.6

koic:
The interface of `ERB.new` will change from Ruby 2.6.

> Add :trim_mode and :eoutvar keyword arguments to ERB.new.
> Now non-keyword arguments other than first one are softly deprecated
> and will be removed when Ruby 2.5 becomes EOL. [Feature #14256]

https://github.com/ruby/ruby/blob/2311087b685e8dc0f21f4a89875f25c22f5c39a9/NEWS#stdlib-updates-outstanding-ones-only

The following address is related Ruby's commit.
https://github.com/ruby/ruby/commit/cc777d0

This PR uses `ERB.version` to switch `ERB.new` interface. Because Thor supports multiple Ruby versions, it need to use the appropriate interface.

Using `ERB.version` instead of `RUBY_VERSON` is based on the following patch.
https://github.com/ruby/ruby/pull/1826

This patch is built into Ruby.
https://github.com/ruby/ruby/commit/40db89c0934c23d7464d47946bb682b9035411f9

### DIFF
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -117,7 +117,13 @@ class Thor
       context = config.delete(:context) || instance_eval("binding")
 
       create_file destination, nil, config do
-        content = CapturableERB.new(::File.binread(source), nil, "-", "@output_buffer").tap do |erb|
+        match = ERB.version.match(/(\d\.\d\.\d)/)
+        capturable_erb = if match && match[1] >= "2.2.0" # Ruby 2.6+
+          CapturableERB.new(::File.binread(source), :trim_mode => "-", :eoutvar => "@output_buffer")
+        else
+          CapturableERB.new(::File.binread(source), nil, "-", "@output_buffer")
+        end
+        content = capturable_erb.tap do |erb|
           erb.filename = source
         end.result(context)
         content = yield(content) if block


yahonda:
I would like this pull request merged since there are a bunch of warnings at Rails CI with Ruby 2.6. https://travis-ci.org/rails/rails/jobs/357183623

There are two failures with Ruby 2.6, then I have opened #595, For 1.8.7 failures at https://travis-ci.org/erikhuda/thor/jobs/357338902 I simply wanted to drop Ruby 1.8.7 support.

koic:
@yahonda Thanks for your investigation. I replaced it with regular expression that you showed 🙏

lib/thor/actions/file_manipulation.rb
```diff
@@ -117,7 +117,13 @@ def template(source, *args, &block)
       context = config.delete(:context) || instance_eval("binding")
 
       create_file destination, nil, config do
-        content = CapturableERB.new(::File.binread(source), nil, "-", "@output_buffer").tap do |erb|
+        match = ERB.version.match(/(\d\.\d.\d)/)
```
yahonda: My previous commit did not have a necessary backslash for the second "." 
```ruby
match = ERB.version.match(/(\d\.\d\.\d)/)
```

Here the dot should be actual dot character not any one character.

koic: I couldn't write a more correctly regular expression because of lack of self review 💦 I fixed it.

yahonda: Thank you for the update. I believe we did our best.

yahonda:
[Rails railties testing against ruby-head (aka Ruby 2.6.0 dev)](https://travis-ci.org/rails/rails/jobs/366998551) has been failing due to
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
```

And the biggest number of warnings will be fixed by this pull request. Here are number of warnings.

```ruby
$ grep 'warning:' log.txt | sort | uniq -c | sort -rn | head -n 5
6092 /home/travis/.rvm/gems/ruby-head/gems/thor-0.20.0/lib/thor/actions/file_manipulation.rb:120: warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.
6092 /home/travis/.rvm/gems/ruby-head/gems/thor-0.20.0/lib/thor/actions/file_manipulation.rb:120: warning: Passing eoutvar with the 4th argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, eoutvar: ...) instead.
5949 /home/travis/.rvm/gems/ruby-head/gems/thor-0.20.0/lib/thor/actions/file_manipulation.rb:120: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
 183 /home/travis/.rvm/gems/ruby-head/gems/sqlite3-1.3.13/lib/sqlite3/pragmas.rb:34: warning: mismatched indentations at 'else' with 'case' at 21
 116 /home/travis/.rvm/gems/ruby-head/gems/sqlite3-1.3.13/lib/sqlite3/pragmas.rb:26: warning: mismatched indentations at 'else' with 'case' at 23
$
```

Thank you.

lib/thor/actions/file_manipulation.rb
```diff
@@ -117,7 +117,13 @@ def template(source, *args, &block)
       context = config.delete(:context) || instance_eval("binding")
 
       create_file destination, nil, config do
-        content = CapturableERB.new(::File.binread(source), nil, "-", "@output_buffer").tap do |erb|
+        match = ERB.version.match(/(\d\.\d\.\d)/)
```
ahorek: This won't work on versions with more numerics. While it's not a problem now, it could cause wierd errors in the future.

example
```ruby
'2.11.0'.match(/(\d\.\d\.\d)/) -> nil
```
an old behaviour is used even if '2.11.0' is newer than '2.2.0'
```ruby
'2.11.0' >= '2.2.0' -> false
```

PR #597, opened 2018-03-28
Accumulate values from multiple switch instances with :hash or :array types

sczizzo:
🌈

Allows multi-valued options (Hash or Array) to be specified multiple times, and all the values are accumulated. To illustrate, suppose you configured a Hash option, e.g.
```
option :env, :type => :hash
```
Prior to this change, you'd pass a single `--env` switch to your executable, e.g.
```
$ my-thing --env FOO:foo BAR:bar
# options => { "env" => { "FOO" => "foo", "BAR" => "bar" } }
```
If you tried to use multiple instances of the `--env` switch, only the last one would take effect:
```
$ my-thing --env FOO:foo --env BAR:bar
# options => { "env" => { "BAR" => "bar" } }
```
After this change, you can still use a single `--env` switch with multiple values:
```
$ my-thing --env FOO:foo BAR:bar
# options => { "env" => { "FOO" => "foo", "BAR" => "bar" } }
```
But now, using multiple `--env` switches will also work as one might expect:
```
$ my-thing --env FOO:foo --env BAR:bar
# options => { "env" => { "FOO" => "foo", "BAR" => "bar" } }
```

### DIFF
--- a/lib/thor/parser/arguments.rb
+++ b/lib/thor/parser/arguments.rb
@@ -39,6 +39,8 @@ class Thor
 
     def parse(args)
       @pile = args.dup
+      @hashes = Hash.new { |h, k| h[k] = {} }
+      @arrays = Hash.new { |h, k| h[k] = [] }
 
       @switches.each do |argument|
         break unless peek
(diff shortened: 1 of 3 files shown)


sczizzo:
Not sure this would ever be merged. AFAICT this repo isn't really under active development; mostly I see bugfixes and the like.

PR #601, opened 2018-05-17
Updates method documentation for ask

pallan:
This is a simple documentation update. I stumbled over the `default` option when I tried to create my own override for `ask` that included it.


rafaelfranca:
Thanks!

PR #609, opened 2018-05-31
Drop old ruby compatibility

deivid-rodriguez:
Hello!

I tried to rebase another PR to get tests passing and its tests failed because `Process.spawn` is not available on ancient rubies.

Maybe it's time for a little housekeeping?


rafaelfranca:
Thank you for the pull request but given bundler uses thor and bundler still support all those rubies we can't drop old rubies support just yet. After the bundler 2.0 release we can think about it.

deivid-rodriguez:
Bundler uses its own vendored `thor`, currently at `0.20`. I assume the reason is that you want to not need to backport anything to 0.x if bundler needed a `0.21` release, but just cut if from master?

deivid-rodriguez:
Anyways, sounds good to me! I'll close this and reopen in the future if necessary. Thanks @rafaelfranca :muscle:

rafaelfranca:
yeah, I'm avoiding have to maintain two release branches.

PR #611, opened 2018-07-05
Add language about how not to use Thor

bosoxbill:
In order to get a clean bill of health from ruby-audit, we should ensure any application developers are aware that they should not just take user input and pass it straight through to Thor.

Might render https://github.com/rubysec/ruby-advisory-db/pull/341 moot?


bosoxbill:
More than happy to do it.

bosoxbill:
What do I do about the travis fails?

PR #616, opened 2018-08-28
Fix print_wrapped to properly parse "\x5" newline character

Choms:
:rainbow: Hi there!

While using Thor I found that the print_wrapped function behaves weirdly when manually inserting \x5 sequence (as per your docs) to force a line break. The original function doesn't takes those line breaks into consideration when calculating the length of the line, and the result is some extra weird line breaks on a long paragraph that includes such sequences (e.g. giving several execution examples).

I've rewrote this function, I reckon it's less "stylish", ~~and has a side effect that is unlikely a print_wrapped will use the whole terminal size~~, but it won't break intended multi-line paragraphs either.

Long story short, I'm breaking a paragraph in words, doing the "\x5" to "\n" replacement and keeping a counter that tells me when the line is over the terminal size, so it places a line break.

This change doesn't interfere with core functionality and I've tested it, feel free to review it and if you see fit, merge (I've also increased the version number).

PS: Both in your "master" branch and my fork I'm getting 4 errors related to Readline spec when running `bundle exec thor spec` locally ¯\\\_(ツ)_/¯

Cheers!

### DIFF
--- a/lib/thor/shell/basic.rb
+++ b/lib/thor/shell/basic.rb
@@ -234,9 +234,9 @@ class Thor
           unwrapped.split(" ").inject do |memo, word|
             word = word.gsub(/\n\005/, "\n").gsub(/\005/, "\n")
             counter = 0 if word.include? "\n"
-            if (counter + word.length + indent) < (width + word.length)
+            if (counter + word.length + 1) < width
               memo = "#{memo} #{word}"
-              counter += word.length+indent
+              counter += (word.length + 1)
             else
               memo = "#{memo}\n#{word}"
               counter = word.length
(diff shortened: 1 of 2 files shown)


Choms:
mmmm I see travis doesn't likes the changes, I'll look into it when I have some time (over the weekend likely) or if anyone else has a hint (it must be related to travis terminal_width - maybe changing the spec to expect an extra newline would do the trick) help is appreciated :)

Choms:
I did some weird calculation with the indentation (for some reason I didn't saw the indent being subtracted already from the terminal width).
I fixed that on the last commit, now it's all passing and actually using the whole terminal width, should be good to merge.

MaxLap:
I can't speak about the details of the change, but the version shouldn't be changed like that as part of the commit. When the maintainer does a release, he will increment it.

rafaelfranca:
You removed the version file instead of only removing the version change. This is why the tests are failing.

Choms:
yea I see feel free to add it merge or just close the PR, I thought it
would be useful, I really don't care about versioning discussions

rafaelfranca:
Sounds good. Can you share at least which string you are using to test so I can write a test to make sure this behavior keep fixed.

rafaelfranca:
Awesome! Thank you so much. I'll merge and add the tests myself.

PR #618, opened 2018-09-10
Fix check_unknown_options! when parsing gets stopped

MaxLap:
If the parsing gets stopped by either a `--` or by `#stop_on_unknown_option!`, then `check_unknown_options!` would still try to do validation on the unparsed parts.

This change now tracks where the stop happened and only check the `@extra` that were before that point.

For example, right now, `exec something -- --verbose` would fail because of `unknown switch --verbose` when `check_unknown_options!` is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Fix `check_unknown_options!` to not check the content that was not parsed, i.e. after a `--` or after the first unknown with `stop_on_unknown_option!`
+
 ## 0.20.0
 * Add `check_default_type!` to check if the default value of an option matches the defined type.
   It removes the warning on usage and gives the command authors the possibility to check for programming errors.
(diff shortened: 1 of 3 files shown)


marcandre:
🎉 Thanks @rafaelfranca for the quick merge.

MaxLap:
Thank you @rafaelfranca. Any plan for a new release? This bug is holding back my switch to Thor.

MaxLap:
@rafaelfranca, did you get a chance to make sure bundler and rails test suite pass?

marcandre:
@rafaelfranca Anything we can do to help? This is blocking our releases of `DeepCover`.

rafaelfranca:
Working on that right now. Sorry for the delay.

rafaelfranca:
0.20.1 released.

marcandre:
:tada: Thanks @rafaelfranca 😸

PR #620, opened 2018-09-11
Fix edge cases in `#relative_to_original_destination_root`

MaxLap:
There are many edge-cases in the existing implementation of `#relative_to_original_destination_root`:

* It does the replacement multiple times
* It can do the replacement in the middle of the received path
* It will do the replacement even if it cuts a directory name (think `/app` versus `/application`)

This is an improvement over #588, as this also solves the same case. (The first commit actually comes from that PR)


MaxLap:
Didn't think to test in 1.8.7. It really requires weird code patterns sometimes.
It's fixed

PR #624, opened 2018-09-25
Why does this make a test fail?

marcandre:
For some reason this outputs "Unknown switches '-w'" even though it is a class_option.

### DIFF
--- a/spec/register_spec.rb
+++ b/spec/register_spec.rb
@@ -1,6 +1,8 @@
 require "helper"
 
 class BoringVendorProvidedCLI < Thor
+  check_unknown_options!
+
   desc "boring", "do boring stuff"
   def boring
     puts "bored. <yawn>"


Choms:
I think `CompatibleWith19Plugin` is missing a line with `default_command :animal` but to be honest I find this whole test spec quite confusing

marcandre:
> I think `CompatibleWith19Plugin` is missing a line with `default_command :animal` but to be honest I find this whole test spec quite confusing

Looks like `Thor::Group` doesn't have that method, only `Thor` has.

In any case, I don't see how being strict about unknown options should have an impact about a class option like in this example.

If this is a bug, it should be fixed before we can consider deprecating the default...

MaxLap:
When registering a Group. All that `Thor` does is define a command on the top-level. In this case, it will add `BoringVendorProvidedCLI#zoo`, which will call invoke on the `CompatibleWith19Plugin`.

This means that `zoo` is a command on the `BoringVendorProvidedCLI`. And because `BoringVendorProvidedCLI` is now `check_unknown_options!`, this is applied to the `zoo` command too.

Maybe adding an actual subcommand would make more sense, but I didn't look into the details of the current handling for subcommands.

This is part of the "invoke" system of Thor which I find hard to find use-case for. But that's just my opinion .

marcandre:
Cleaning up my open PRs. Closing this as not much feedback

PR #625, opened 2018-09-25
Deprecate relying on default exit_on_failure?

marcandre:
This forces implementers in defining `exit_on_failure?`, or else a deprecation warning is issued on error. See discussion in #621


marcandre:
Mm, my branch was broken, sorry for that. Should be fixed now.

PR #626, opened 2018-09-25
Deprecate default check

marcandre:
This issues a deprecation warning if one specifies an incompatible default value without either calling `allow_incompatible_default_type!` or passing `check_default_type: false`.

See discussion in #621

### DIFF
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -111,7 +111,7 @@ class Thor
 
     def validate!
       raise ArgumentError, "An option cannot be boolean and required." if boolean? && required?
-      validate_default_type! if @check_default_type
+      validate_default_type!
     end
 
     def validate_default_type!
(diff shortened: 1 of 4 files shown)


lib/thor/parser/option.rb
```diff
@@ -127,8 +127,15 @@ def validate_default_type!
...
       end
-
-      raise ArgumentError, "Expected #{@type} default value for '#{switch_name}'; got #{@default.inspect} (#{default_type})" unless default_type == @type
+      if default_type != @type
+        err = "Expected #{@type} default value for '#{switch_name}'; got #{@default.inspect} (#{default_type})"
+        if @check_default_type
+          raise ArgumentError, err
+        elsif @check_default_type == nil
+          Thor.deprecation_warning "#{err}.\n" +
+            'This will be rejected in the future unless you explicitly pass the options `check_default_type: false`.'
```
Choms: I'd rephrase that as:
```'This will be rejected in the future unless you explicitly pass the options `check_default_type: false` or define `allow_incompatible_default_type!` in your code'```

File: spec/thor_spec.rb [hunk 2]
```diff
@@ -694,12 +694,32 @@ def unknown(*args)
...
+
+          option "bar", :type => :numeric, :default => "foo"
+        end
+      end.not_to output.to_stderr
+    end
+
+    it "allows incompatible types if `check_default_type: false` is given" do
+      expect do
+        Class.new(Thor) do
+          allow_incompatible_default_type!
```
Choms: Remove this line so the second test makes sense (else it's equivalent to the previous one)

marcandre: Oops, indeed. And it revealed a bug that was already there whereby an explicit setting was always overwritten by a merge being inverted. Fixed

lib/thor/parser/option.rb
marcandre: As above, but "or *call*allow_..."

PR #628, opened 2018-10-23
Abort when `run` subprocess fail

deivid-rodriguez:
Hi!

Sometimes my Rails generators that spawn subprocesses crash, and I don't even notice because my build status stays green. I'd like to add a "safe" mode to `run` that when passed makes the process abort if the subcommand fails.


deivid-rodriguez:
Thanks so much @rafaelfranca!

PR #630, opened 2018-10-23
Support did-you-mean functionality in thor

kddeisz:
When an invocation errors out because it cannot find a corresponding command, this will attempt to suggest alternatives in the case of typos. Also, when invalid switches are passed and checking for invalid switches is enabled, it will attempt to suggest alternatives as well.


kddeisz:
Any interest in this? Trying to get this merged so bundler can take advantage

PR #636, opened 2018-11-10
Remove the globally installed gem by rvm

y-yagi:
This fixes follow error:

```
Gem::LoadError: You have already activated did_you_mean 1.2.0, but your Gemfile requires did_you_mean 1.2.1. Prepending `bundle exec` to your command may solve this.
```


y-yagi:
cc @rafaelfranca

PR #637, opened 2018-11-10
Add care about old version of `did_you_mean`

y-yagi:
`DidYouMean::SpellChecker` class added by `v1.0.2`.
https://github.com/yuki24/did_you_mean/blob/master/CHANGELOG.md#v102

Therefore, even if can use `did_you_mean`, may not be able to use
`DidYouMean::SpellChecker`. For that case, also ignored `NameError`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,8 @@ group :test do
   else
     gem "webmock"
   end
-  if RUBY_VERSION >= '2.5.0'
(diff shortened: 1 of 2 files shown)


y-yagi:
cc @rafaelfranca

mathieujobin:
Anyone tested if this fix bundler-audit?

PR #643, opened 2018-12-30
TravisCI: removed Ruby versions < 2.3 since rubygems deprecated them

alessio-signorini:
```ruby
$ gem update --system
Updating rubygems-update
Fetching: rubygems-update-3.0.1.gem (100%)
ERROR:  Error installing rubygems-update:
	rubygems-update requires Ruby version >= 2.3.0.
ERROR:  While executing gem ... (NoMethodError)
    undefined method `version' for nil:NilClass
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ before_install:
 bundler_args: --without development
 language: ruby
 rvm:
-  - 1.8.7
-  - 1.9.3
-  - 2.0.0
-  - 2.1.10
-  - 2.2.10
   - 2.3.7
   - 2.4.4
   - 2.5.1


rafaelfranca:
Where can I read more about this removal? Thor still support Ruby < 2.3 so we need to test on them in some way

rafaelfranca:
Right. An easier fix is to change the `gem update --system` call on travis to define a version that is not rubygems 3.0 for rubies < 2.3.

alessio-signorini:
@rafaelfranca that's an idea. I looked at [this link](https://docs.travis-ci.com/user/languages/ruby/#bundler-20) but could not find an easy way to do it. Do you know how? Do you mind creating a PR for that? How do we know the Ruby versions that need to be supported by Thor?

piotrmurach:
As far as I understand each ruby installation includes a version of rubygems. Even the latest Ruby `v2.6.0` isn't released with rubygems `v3.0`. Therefore this is an issue for future releases of Ruby. Therefore I would be inclined to remove `gem update --system` from the travis config file all together. I would also constraint `gem install bundler` to version `< 2.0.0` due to imcompatibility with rubies `< 2.3.0`.

deivid-rodriguez:
Maybe we can actually drop support instead. Thor's main dependants, Rails and Bundler, have already moved on, so maybe it's time?

piotrmurach:
@deivid-rodriguez That sounds like a drastic step. I do use thor in my project called [tty](https://github.com/piotrmurach/tty) and I'm keen for this library to support Ruby `>= 2.0.0`. I belive there is value in maintaining older versions of Ruby and progressively eliminating more defunct ones. For example, dropping `1.8.7` support could be a good first step?

deivid-rodriguez:
My take to this is to follow the upstream policy, so when ruby-core drops support for a version of ruby, I drop support too. But this is up to the maintainers to decide :)

rafaelfranca:
I'm up to remove support of older version on thor 1.0 but right now I'd prefer a green build in all the versions we support today.

deivid-rodriguez:
So that means that the next version will not yet be 1.0, or at least that you haven't yet decided it... Right?

deivid-rodriguez:
What about for dropping support for rubies under 2.0, @rafaelfranca? Would you be willing to merge that now if I open a PR?

PR #649, opened 2019-03-04
Travis CI refresh

deivid-rodriguez:
Since it looks like #643 is not going to happen any time soon, I'd like to a least get CI green.


deivid-rodriguez:
Updated to pick up the latest ruby and the latest rubygems security release.

PR #651, opened 2019-03-13
Abort if the command for `run` fails and `exit_on_failure?` is true

hibariya:
Hi, there :rainbow:

I'd like to change the error handling of `Thor::Actions#run` for the following reason. 

According to the description of `exit_on_failure?`, the process should exit with status 1 if any error happens. I think that it should be also applied to `run`.

Although we can pass `abort_on_failure` option for `run`, passing it every time seems a little redundant when we call `run` multiple times. This is the other reason for this change.

For instance, when an error occres [here](https://github.com/rails/webpacker/blob/e95dd724ea3a385b9cc6935c3a07a34f1d1f9657/lib/install/elm.rb#L27), I want to notice not only by the output but also by the status code.

### DIFF
--- a/lib/thor/actions.rb
+++ b/lib/thor/actions.rb
@@ -259,7 +259,7 @@ class Thor
 
       result = config[:capture] ? `#{command}` : system(command.to_s)
 
-      if config[:abort_on_failure]
+      if config.fetch(:abort_on_failure, self.class.exit_on_failure?)
         success = config[:capture] ? $?.success? : result
         abort unless success
       end
--- a/lib/thor/base.rb
+++ b/lib/thor/base.rb
@@ -506,6 +506,11 @@ class Thor
         raise InvocationError, msg
       end
 
+      # A flag that makes the process exit with status 1 if any error happens.
+      def exit_on_failure?
+        false
+      end
+
     protected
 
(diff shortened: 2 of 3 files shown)


hibariya:
#649 is needed to fix the error on CI. A build with that patch seems work.
https://travis-ci.org/hibariya/thor/builds/505774428

deivid-rodriguez:
I really like this idea! :heart:

Even after adding the `abort_on_failure:` flag I still miss errors in my generators, because I cannot control every call to `Thor::Actions#run`.

For example, I just noticed that our Rails 6 application generation in ActiveAdmin is [crashing](https://circleci.com/gh/activeadmin/activeadmin/13190), but the crash is only buried in the logs, so there's no way to notice other than manual log scanning (or if the failure affects further tests, but in that case, the culprit is still very hidden).

hibariya:
I've rebased onto the top of the master. The build seems OK.

deivid-rodriguez:
Any chance on getting a review here?

kaspth:
👍, this also matches the original intent https://github.com/erikhuda/thor/pull/628

I don't have commit access, so it'll have to be @rafaelfranca who merges ❤️

PR #652, opened 2019-03-27
Multiple usages

deivid-rodriguez:
On bundler, we have some commands with quite complicated synopsis. For example, `bundle update`, which has different valid combinations of arguments and command line flags, pretty hard to expresses on a single usage line without making it look like an understandable mess.

Instead, I would like to be able to specify multiple synopsis, so that `thor` would give a error of the following type when an invalid command line is specified:

```
ERROR: "bundle update" was called with no arguments
Usage: bundle update *GEMS [OPTIONS]
       bundle update --group *GROUPS [OPTIONS]
       bundle update --source *SOURCES [OPTIONS]
       bundle update --ruby
       bundle update --bundler
       bundle update --all
```

In general, this is useful for mutually exclusive combinations of flags and arguments.
    
This is explained in http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html, bullet point 8, and it's a technique used by many CLI utils.
    
For example, the following is the beginning of the man's man page:
    
    NAME
           man - an interface to the on-line reference manuals
    
    SYNOPSIS
           man  [-C  file] [-d] [-D] [--warnings[=warnings]] [-R encoding] [-L locale] [-m system[,...]] [-M path] [-S list] [-e extension] [-i|-I] [--regex|--wildcard]
           [--names-only] [-a] [-u] [--no-subpages] [-P pager] [-r prompt] [-7] [-E encoding] [--no-hyphenation]  [--no-justification]  [-p  string]  [-t]  [-T[device]]
           [-H[browser]] [-X[dpi]] [-Z] [[section] page[.section] ...] ...
           man -k [apropos options] regexp ...
           man -K [-w|-W] [-S list] [-i|-I] [--regex] [section] term ...
           man -f [whatis options] page ...
           man  -l  [-C  file]  [-d]  [-D]  [--warnings[=warnings]]  [-R  encoding]  [-L locale] [-P pager] [-r prompt] [-7] [-E encoding] [-p string] [-t] [-T[device]]
           [-H[browser]] [-X[dpi]] [-Z] file ...
           man -w|-W [-C file] [-d] [-D] page ...
           man -c [-C file] [-d] [-D] page ...
           man [-?V]

    
This commit implements support for this kind of synopsis, by allowing passing an array of "usages" instead of a single one.


deivid-rodriguez:
Thoughts about this?

deivid-rodriguez:
Friendly ping :)

deivid-rodriguez:
I'm getting ready in bundler for an eventual thor release, and I see that this PR seems to have caused some issues. I'll investigate those and provide a follow-up as soon as possible :+1:.

deivid-rodriguez:
Actually, nevermind, it's #679 causing problems. I'll investigate.

PR #656, opened 2019-05-10
utf8 identical checking

chenkovsky:
File.binread returns string with ASCII-8BIT encoding. but the rendered result is Utf8 encoding. For non-ascii characters, although the content is same, the comparison will return false. when we invoke migration generate with force option. it will delete the old migration file, and create a new migration file with the same content.

### DIFF
--- a/lib/thor/actions/create_file.rb
+++ b/lib/thor/actions/create_file.rb
@@ -43,7 +43,7 @@ class Thor
       # Boolean:: true if it is identical, false otherwise.
       #
       def identical?
-        exists? && File.binread(destination) == render
+        exists? && File.binread(destination) == String.new(render).force_encoding("ASCII-8BIT")
       end
 
       # Holds the content to be added to the file.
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@chenkovsky Can we add a test for this change, so it doesn't break again in the future?

chenkovsky:
@deivid-rodriguez I change the test. add utf8 character to content.

deivid-rodriguez:
Sorry for dropping the ball here.

For the test, I would personally prefer a new separate test that uses UTF-8 content, rather that changing every test to create a file with UTF-8 content.

Other than that this PR looks good to me.

Nytorian:
Heyy, just a little suggestion from me:

How about introducing a 'wrapper' method for `File.binread(destination)` and similar that output the earlier mentioned `ASCII-8BIT`? Ideally, we could force the output of `binread` to `Encoding.default_external` which would be more flexible for sure.

Regards

rafaelfranca:
Closed by #786

PR #664, opened 2019-06-26
Drop ruby 1.x support

deivid-rodriguez:
References https://github.com/erikhuda/thor/pull/643#issuecomment-490611243.


deivid-rodriguez:
Thank you! :heart_eyes:

PR #668, opened 2019-07-24
Prefer `require_relative` for internal requires

deivid-rodriguez:
I'd like to use `require_relative` for internal requires now that ruby 1.8 support has been dropped. It should be faster and it makes it more clear that the file being required is in the same library.

Also, it's nice to do this for `bundler` (which vendors a copy of `thor`) because given the `LOAD_PATH` manipulation `bundler` does, and that a copy of bundler is now always present in the `LOAD_PATH` by default in every installation of ruby, `require_relative` is safer because we are sure we never end up requiring a different copy of bundler than the one we're currently running.


deivid-rodriguez:
Thank you so much @rafaelfranca :)

PR #677, opened 2019-09-15
Suppress `URI.open` warning in Ruby 2.7

koic:
This PR suppresses the follwoing warning in Ruby 2.7

```console
% ruby -v
ruby 2.7.0dev (2019-09-14T09:21:37Z master 39c37acf86) [x86_64-darwin17]
% bundle exec thor spec
(snip)

/Users/koic/src/github.com/erikhuda/thor/lib/ty
./Users/koic/src/github.com/erikhuda/thor/lib/thor/actions/file_manipulation.rb:89:
warning: calling URI.open via Kernel#open is deprecated, call URI.open directly
```

cf https://github.com/ruby/ruby/commit/05aac90a1bcfeb180f5e78ea8b00a4d1b04d5eed.

### DIFF
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -80,13 +80,13 @@ class Thor
       config = args.last.is_a?(Hash) ? args.pop : {}
       destination = args.first
 
-      if source =~ %r{^https?\://}
+      render = if source =~ %r{^https?\://}
         require "open-uri"
+        URI.send(:open, source) { |input| input.binmode.read }
       else
         source = File.expand_path(find_in_source_paths(source.to_s))
+        open(source) { |input| input.binmode.read }
       end
-
-      render = open(source) { |input| input.binmode.read }
 
       destination ||= if block_given?
         block.arity == 1 ? yield(render) : yield


lib/thor/actions/file_manipulation.rb
```diff
@@ -80,14 +80,14 @@ def get(source, *args, &block)
       config = args.last.is_a?(Hash) ? args.pop : {}
       destination = args.first
 
-      if source =~ %r{^https?\://}
+      render = if source =~ %r{^https?\://}
         require "open-uri"
+        URI.send(:open, source) { |input| input.binmode.read }
```
rafaelfranca: Why we need the send?

koic: It uses `send` method because this is a private method in Ruby 2.4 and lower. 

## Ruby 2.5

```console
% ruby -ropen-uri -ve 'URI.open("https://github.com/erikhuda/thor")'
ruby 2.5.6p201 (2019-08-28 revision 67796) [x86_64-darwin17]
```

## Ruby 2.4

```console
% ruby -ropen-uri -ve 'URI.open("https://github.com/erikhuda/thor")'
ruby 2.4.7p357 (2019-08-28 revision 67796) [x86_64-darwin17]
-e:1:in `<main>': private method `open' called for URI:Module (NoMethodError)
```

PR #679, opened 2019-09-23
Fix `--skip-foo` switches when there's a trailing non-switch

### LINKED ISSUE #580: boolean options changed in v0.20.0
I think I have found a possible bug in boolean options.  The value of the option is different when a CLI argument is included.

```ruby
        desc 'test1', 'This is test1 CLI'
        option :skip_upload, default: false, type: :boolean
        def test1
          puts "test1: skip=#{options[:skip_upload]}"
        end

        desc 'test2', 'This is the test2 CLI'
        option :skip_upload, default: false, type: :boolean
        def test2(param)
          puts "test2: skip=#{options[:skip_upload]}"
        end
```

```
$ bundle exec exe/astcli cdr test1
test1: skip=false
$ bundle exec exe/astcli cdr test2 argument
test2: skip=false

$ bundle exec exe/astcli cdr test1 --skip-upload
test1: skip=true
$ bundle exec exe/astcli cdr test2 --skip-upload argument
test2: skip=false
```

deivid-rodriguez:
I was running  the command `bundle exec rails new --dev --skip-webpack-install my_project` inside my Rails clone, but Rails generators kept running `--skip-webpack-install`.

Turns out `thor` was considering `my_project` as the value for the `--skip-webpack-install` switch. However, I think the command line I specified is correct and non-ambiguous and that `--no-` and `--skip-` flags are not supposed to have a value.

This PR fixes this edge case.

It adds to tests, the first one is already passing against master, the second one is fixed by this PR. So, I think it also makes the behaviour more consistent.

Fixes #580.

### DIFF
--- a/lib/thor/parser/options.rb
+++ b/lib/thor/parser/options.rb
@@ -196,7 +196,10 @@ class Thor
     # Parse boolean values which can be given as --foo=true, --foo or --no-foo.
     #
     def parse_boolean(switch)
-      if current_is_value?
+      no_or_skip = no_or_skip?(switch)
+      if no_or_skip
+        @switches.key?(switch) || !no_or_skip
+      elsif current_is_value?
         if ["true", "TRUE", "t", "T", true].include?(peek)
           shift
           true
@@ -204,10 +207,10 @@ class Thor
           shift
           false
         else
-          !no_or_skip?(switch)
+          !no_or_skip
         end
       else
-        @switches.key?(switch) || !no_or_skip?(switch)
+        @switches.key?(switch) || !no_or_skip
       end
     end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I suspected that more people would've run into this and I found an open issue: https://github.com/erikhuda/thor/issues/580. I'll edit the PR description so it gets autoclosed if/when merged.

doudou:
Thanks for fixing this ... I completely forgot about it, got used to always specify the boolean flags with `=f` instead (`--webpack-install=f`).

deivid-rodriguez:
No problem!

deivid-rodriguez:
@doudou Funny enough, my fix broke your workaround and I got beaten by it myself. I added [a follow up PR](https://github.com/erikhuda/thor/pull/682) to restore support for that syntax so that you and I don't have to change all of our scripts, tests or whatever to use the previous syntax.

PR #691, opened 2019-11-15
 🌈 More accurate Arguments#current_is_value? check

timothysmith0609:
### Motivation

For a number of tools, the argument `-` is often treated to mean STDIN. Currently, Thor is unable to accept `-` as an argument _value_, since it defines a flag as any entry beginning with `-`. This seems overly restrictive. I would argue that a flag is defined as follows:

>A flag is any entry that begins with 1 or 2 dashes, followed by _at least_ one character

In regex terms, this is `/-{1,2}\S+/`.

### TODO 
- [x] update/create tests
- [x] update docs, if necessary _(edit: does not appear necessary to update any docs)_

### DIFF
--- a/spec/parser/arguments_spec.rb
+++ b/spec/parser/arguments_spec.rb
@@ -40,6 +40,12 @@ describe Thor::Arguments do
       expect(parse("product", "title", "age")["array"]).to eq(%w(title age))
     end
 
+    it "accepts - as an array argument" do
+      create :array => nil
+      expect(parse("-")["array"]).to eq(%w(-))
+      expect(parse("-", "title", "-")["array"]).to eq(%w(- title -))
+    end
+
     describe "with no inputs" do
       it "and no arguments returns an empty hash" do
         create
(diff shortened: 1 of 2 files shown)


rafaelfranca:
This makes sense. Can you write the tests?

spec/parser/arguments_spec.rb
```diff
@@ -40,6 +40,12 @@ def parse(*args)
       expect(parse("product", "title", "age")["array"]).to eq(%w(title age))
     end
 
+    it "accepts - as an array argument" do
+      create :array => nil
+      expect(parse("-")["array"]).to eq(%w(-))
+      expect(parse("-", "title", "-")["array"]).to eq(%w(- title -))
+    end
```
timothysmith0609: I ran this test against master and, as expected, it fails. I don't think there's any other cases we need to consider for this feature.

I also went through the docs and feel like we don't need to update/change anything: IMO this change is to allow functionality that should already (implicitly) be present, not a new feature.

PR #692, opened 2019-11-15
Add :env option to Actions#run

jonathanhefner:
This option sets environmental variables for the executed command.

---

@deivid-rodriguez  Is this what you had in mind in https://github.com/rails/rails/pull/34980#discussion_r250640653?


deivid-rodriguez:
Yeah! That was it! Thank you :purple_heart:

PR #697, opened 2019-12-11
Allow for nested no_commands blocks.

TomFinill:
### Background
In certain circumstances we may wish to nest `no_commands` blocks.

However, since when exiting the `no_commands` block the `@no_commands`
instance var is reset to `false`, we get unexpected behaviour.

For example, this will show warnings stating that the method `foo` is missing usage and a description:

```ruby
class Foo < Thor

  no_method do
    attr_reader :bar

    def foo
      'foo!'
    end
  end
end
```

This particular example is trivially solved but things get more difficult when, say, including modules.

```ruby
# lib/thor/base.rb
def no_commands
  @no_commands = true
  yield
ensure
   @no_commands = false
end
```

### Solution

This change uses a `NestedContext` object to track the depth of the `no_command` blocks and ensure that we only start creating commands again once we've left all of them.

### Comments

- I'm definitely not sold on the name `NestedContext`

### DIFF
--- a/lib/thor/base.rb
+++ b/lib/thor/base.rb
@@ -2,6 +2,7 @@ require_relative "command"
 require_relative "core_ext/hash_with_indifferent_access"
 require_relative "error"
 require_relative "invocation"
+require_relative "nested_context"
 require_relative "parser"
 require_relative "shell"
 require_relative "line_editor"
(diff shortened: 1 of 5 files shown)


rafaelfranca:
Thank you for the pull request. I think this solution is too complex just to allow nested `no_commands`. Can you explain why you need nested `no_commands`?

TomFinill:
Hello, I have been working on refactoring some scripts for a Rails application and wanted to make use of the `ActiveModel` validation API.  So I included the `ActiveModel::Validations` module which created the nested `no_commands` via an `attr_reader`.

PR #698, opened 2019-12-14
Thor::VERSION should be 1.0.0 now

deivid-rodriguez:


deivid-rodriguez:
Also, could the 1.0.0 release be pushed to github releases, or at least the `v1.0.0` tag created? We use the https://github.com/erikhuda/thor/archive/v1.0.0.zip address to vendor `thor` inside `bundler`, but it doesn't yet exist.

Issue #707, opened 2020-01-10
inject_into_file with pretend reports file unchanged


scambra:
Before this commit 25dc317f71f5d212d63fbd9eb4b81bd437ee762f inject_into_file with pretend option was printing inject line, but now it complains about file didn't change.

The best way it would try to replace and return success, without writing file

dorner:
@scambra not sure if you're still interested in this - it looks good to me but there should be a spec to cover this behavior.

scambra:
yes, I'm still interested, I will look into specs whenever I can

scambra:
@dorner I have added a spec for inject_into_file with pretend

PR #708, opened 2020-01-10
Fix for #707

scambra:
Try to replace text but skip overwritting file with pretend option


dorner (approved):
Looks good to me! @rafaelfranca

PR #712, opened 2020-01-28
🌈 Update `Thor::Actions#inside` to return the value yielded by the block

jordan-brough:
🌈
So that calling code can more conveniently inspect & use this value.

The value previously returned by this method doesn't seem useful so this seems like a safe change to make.

If this looks OK, should I also update the CHANGELOG?


jordan-brough:
@rafaelfranca any thoughts on this?

dorner (approved):
Looks good to me!

PR #714, opened 2020-02-21
Indent multiline messages in say_status

jonathanhefner:
This prevents multiline status messages from breaking the left margin.

As an example, before this commit:

```
     status1  single line
     status2  multiline line 1
multiline line 2
multiline line 3
     status3  multiline indent 0
  multiline indent 2
    multiline indent 4
     status4  single line
```

And after this commit:

```
     status1  single line
     status2  multiline line 1
              multiline line 2
              multiline line 3
     status3  multiline indent 0
                multiline indent 2
                  multiline indent 4
     status4  single line
```

Closes rails/rails#38387.

### DIFF
--- a/lib/thor/shell/basic.rb
+++ b/lib/thor/shell/basic.rb
@@ -109,13 +109,14 @@ class Thor
       def say_status(status, message, log_status = true)
         return if quiet? || log_status == false
         spaces = "  " * (padding + 1)
-        color  = log_status.is_a?(Symbol) ? log_status : :green
-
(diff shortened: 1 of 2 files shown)


alexpapworth:
@rafaelfranca any chance we can merge this one?

dorner:
I like this change but it might be breaking behavior since existing usage may be compensating for this by adding more spaces. @rafaelfranca is this something we should leave for a major version bump? Or add a flag for this behavior as an argument?

PR #715, opened 2020-02-27
🌈 Need to dup default or else we have a troublesome alias

tsontario:
Following work on https://github.com/Shopify/krane/issues/697, I discovered a subtle bug that is now causing issues with the [repeatable flag feature](https://github.com/erikhuda/thor/pull/674).

If a default argument for a `Thor::Option | Thor::Argument` is provided, the option/argument constructor will call:
```
@assigns[argument.human_name] = argument.default
```
This aliases, __and does not create a copy of__, `argument.default` to `@assigns[argument.human_name]`. So now every time we update `@assigns`, we are also mutating `argument.default`. This was never a problem before because, without a repeatable flag, we'd never encounter a failure mode.

#### TODO
- [x] unit test

### DIFF
--- a/lib/thor/parser/arguments.rb
+++ b/lib/thor/parser/arguments.rb
@@ -30,7 +30,12 @@ class Thor
 
       arguments.each do |argument|
         if !argument.default.nil?
-          @assigns[argument.human_name] = argument.default
+          begin
+            @assigns[argument.human_name] = argument.default.dup
+          rescue TypeError  # Compatibility shim for un-dup-able Fixnum in Ruby < 2.4
+            @assigns[argument.human_name] = argument.default
+            next
+          end
         elsif argument.required?
           @non_assigned_required << argument
         end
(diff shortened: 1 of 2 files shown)


lib/thor/parser/arguments.rb
```diff
@@ -30,7 +30,12 @@ def initialize(arguments = [])
 
       arguments.each do |argument|
         if !argument.default.nil?
-          @assigns[argument.human_name] = argument.default
+          begin
+            @assigns[argument.human_name] = argument.default.dup
+          rescue TypeError  # Compatibility shim for un-dup-able Fixnum in Ruby < 2.4
+            @assigns[argument.human_name] = argument.default
+            next
```
tsontario: Before Ruby 2.4, attempting to dup a Fixnum results in a TypeError. From what I can see, that's the only accepted type for an option that could possibly throw, so for the sake of backward's compatibility I've had to add this in

rafaelfranca: So we need to call `next`?

timothysmith0609: Ah, you're right we don't. I'll remove

PR #717, opened 2020-03-05
🌈 Fix readline specs

bpo:
An error was introduced in commit 7061b06fd87dc8b952f9072507ee991ee0a0858e
where readline specs fail because `require "readline"` is called after the
spec has established a mock `::Readline`. Prior to that change
`require "readline"` was always only called once.

This patch:
  * memoizes the `require` to ensure that it only happens once
  * leaves the `Object.const_defined?` check in place to keep tests simple
  * eagerly calls `available?` in the readline spec to match the original
    author's intent by requiring `readline` before the specs start.

This error wasn't caught by CI due to the bundler cache. You can see the failure caused by the earlier change by [removing the Bundler cache from Travis](https://travis-ci.org/bpo/thor/jobs/658865414)

### DIFF
--- a/lib/thor/line_editor/readline.rb
+++ b/lib/thor/line_editor/readline.rb
@@ -2,8 +2,8 @@ class Thor
   module LineEditor
     class Readline < Basic
       def self.available?
-        @readline_reqd ||= begin
-            require "readline"
+        begin
+          require "readline"
         rescue LoadError
         end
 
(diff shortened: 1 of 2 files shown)


lib/thor/line_editor/readline.rb
```diff
@@ -2,8 +2,8 @@ class Thor
   module LineEditor
     class Readline < Basic
       def self.available?
-        begin
-          require "readline"
```
rafaelfranca: `require` by default only happens once. Why do we need this memoization?

bpo: I had myself convinced that subsequent calls after building the mock would raise a `TypeError` here, but I was incorrect. Will update this PR.

PR #720, opened 2020-03-25
`remove_file` should unlink broken symlinks

2called-chaos:
The problem is that `remove_file` only checks if `File.exist?` which is false if the symlink target does not exist (anymore). In [`create_link`](https://github.com/erikhuda/thor/blob/34df888d721ecaa8cf0cea97d51dc6c388002742/lib/thor/actions/create_link.rb#L56) this was already considered.


dorner (approved):
Looks good!

PR #730, opened 2020-06-23
Use string interpolation for trailing whitespace

jonathanhefner:
Code editors are commonly configured to strip trailing whitespace when saving.  Additionally, trailing whitespace may be hard to distinguish visually, both in the editor and in diffs.

This commit uses string interpolation to encode expected trailing whitespace.


dorner (approved):
Looks good!

PR #735, opened 2020-08-25
Check for duplicate content in relevant section when inserting into files

excid3:
I ran into some unexpected behavior today with `insert_into_file`. I was trying to insert `, null: false` into a migration after some text, but nothing happened. 

```ruby
  t.belongs_to :recipient, null: false
  t.string :type
````

```ruby
insert_into_file path, ", null: false", after: "t.string :type"
```

This does nothing because `, null: false` exists in the file (anywhere).

Using `after:` I would expect `insert_into_file` to only care if the string exists _after_ the matching line, not before it.

This PR would change the behavior to only look through the before or after portions of the file content to make it more intuitive.

### DIFF
--- a/lib/thor/actions/inject_into_file.rb
+++ b/lib/thor/actions/inject_into_file.rb
@@ -108,7 +108,10 @@ class Thor
       def replace!(regexp, string, force)
         return if pretend?
         content = File.read(destination)
-        if force || !content.include?(replacement)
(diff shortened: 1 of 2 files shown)


rafaelfranca:
I'll have to revert this PR since it changes the behavior when the regexp contains capture. `split` returns the capture inside the array of split strings, so we can't know with the current implementation what is the after and before the regexp.

PR #737, opened 2020-09-02
Allow leading hyphen in switch values when specified with =

univerio:
This is a fixed version of #498.

This allows `--foo=-bar` to work.

The implementation works by keeping track of an additional piece of state `@force_current_to_be_value`, which is maintained inside `shift` and `unshift` to ensure it matches the state in `@pile`.

### DIFF
--- a/lib/thor/parser/options.rb
+++ b/lib/thor/parser/options.rb
@@ -45,7 +45,7 @@ class Thor
       @switches = {}
       @extra = []
       @stopped_parsing_after_extra_index = nil
-      @force_current_to_be_value = false
+      @is_switch_value = false
 
       options.each do |option|
         @switches[option.switch_name] = option
@@ -76,18 +76,18 @@ class Thor
     end
 
     def shift
-      @force_current_to_be_value = false
+      @is_switch_value = false
       super
     end
 
     def unshift(arg, force_value: false)
-      @force_current_to_be_value = force_value
+      @is_switch_value = force_value
       super(arg)
     end
 
     def parse(args) # rubocop:disable MethodLength
       @pile = args.dup
-      @force_current_to_be_value = false
+      @is_switch_value = false
       @parsing_options = true
 
       while peek
@@ -163,7 +163,7 @@ class Thor
     # Two booleans are returned.  The first is true if the current value
     # starts with a hyphen; the second is true if it is a registered switch.
     def current_is_switch?
-      return [false, false] if @force_current_to_be_value
+      return [false, false] if @is_switch_value
       case peek
       when LONG_RE, SHORT_RE, EQ_RE, SHORT_NUM
         [true, switch?($1)]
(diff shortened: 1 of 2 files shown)


spec/parser/options_spec.rb
```diff
@@ -425,6 +428,7 @@ def remaining
 
       it "accepts a switch=<value> assignment" do
         expect(parse("--attributes=a", "b", "c")["attributes"]).to eq(%w(a b c))
+        expect(parse("--attributes=-a", "b", "-c")["attributes"]).to eq(%w(-a b))
```
dorner, line 431: This seems confusing... if "-a" and "-c" are valid values for an array attribute, then why would both "-a b" and "-a b -c" give you two values?

univerio: The alternative is more confusing IMO and breaks backwards compatibility, i.e. `--attributes=a b -c` parsing to `{"attributes" => ['a', 'b', '-c']}` means that you can no longer have any options at all after an array option, which is not what the user expects.

This is the same reason this PR is limiting its scope to require a `=` to trigger this "leading `-` is a valid value" behavior.

dorner: OK - I think we'd need to document this specifically then.

There seems to be another PR to fix the same issue? https://github.com/erikhuda/thor/pull/498 What's the difference?

univerio: Yes, this aims to fix the same issue as #498, but #498 appears to have been abandoned. I believe it also has a bug as described by [this comment by @rafaelfranca](https://github.com/erikhuda/thor/pull/498#discussion_r100716476):

> I think setting this state here will broke it if there are more options after the first with - in the value, can you add tests to make sure it will work?

univerio: Can you clarify what you're referring to by "option/options portion"? Is this something I should be adding as part of this PR?

lib/thor/parser/options.rb
```diff
@@ -45,6 +45,7 @@ def initialize(hash_options = {}, defaults = {}, stop_on_unknown = false, disabl
       @switches = {}
       @extra = []
       @stopped_parsing_after_extra_index = nil
+      @is_switch_value = false
```
dorner, line 48: This is actually incorrect, no? This is a value with a dash that is specifically *not* a switch.

dorner (changes requested):

spec/parser/options_spec.rb
dorner: My bad - I forgot what project I was talking about 😛  The update would be to the Wiki once this is merged.

lib/thor/parser/options.rb
univerio: I'm not sure I understand. This instance variable is `true` if and only if the current element in the pile being processed is considered to be a value _even if_ it has a leading `-`. The initial state is that nothing has been parsed yet, in which case we want to parse it as an option if it has a leading `-`, not as a value.

dorner: Yep, exactly. So when this variable is true, the current argument is a value, *not* a switch. The name "is switch value" seems to say "the current value is actually a switch".

Maybe something like `is_literal_value` or something similar?

univerio: I landed on `is_treated_as_value`. PTAL @dorner

dorner (approved):
Much better - looks good to me! @rafaelfranca ?

Issue #398, opened 2014-01-17
long_desc: force keeping spaces and not break lines


mdesantis:
Hello,

I have a `long_desc` method like this:

``` ruby
long_desc <<-LONGDESC.gsub("\n", "\x5")
  Examples:

  bin/ruby-rails-documentations                             \\
    generate ~/ruby-rails-documentations                    \\
    --sdoc-dir=~/sdoc --ruby-dir=~/ruby --rails-dir=~/rails \\
    --ruby-versions=1.9.3p484 2.0.0p353 2.1.0               \\
    --rails-versions=3.0.20 3.1.12 3.2.16 4.0.2 4.1.0.beta1
LONGDESC
def generate(output_dir)
  ...
end
```

Executing `ruby-rails-documentations help generate` I want my output to be like this:

``` sh
Examples:

bin/ruby-rails-documentations                             \
  generate ~/ruby-rails-documentations                    \
  --sdoc-dir=~/sdoc --ruby-dir=~/ruby --rails-dir=~/rails \
  --ruby-versions=1.9.3p484 2.0.0p353 2.1.0               \
  --rails-versions=3.0.20 3.1.12 3.2.16 4.0.2 4.1.0.beta1
```

Instead, I get this:

``` sh
Examples:

bin/ruby-rails-documentations \
 generate ~/ruby-rails-documentations \
 --sdoc-dir=~/sdoc --ruby-dir=~/ruby --rails-dir=~/rails \
 --ruby-versions=1.9.3p484 2.0.0p353 2.1.0 \
 --rails-versions=3.0.20 3.1.12 
3.2.16 4.0.2 4.1.0.beta1
```

The merged spaces is not a big problem, but the added newlines is - they break the command syntax, and I can't predict where they will be.

Is there any way to force keeping spaces and, above all, force to not break lines?

jinhianlee:
+1 for @bradland's comment.

ImaginateWayne:
Is this change ever getting merged into master?

hfossli:
Bump

jsuwo:
+1

vigo:
+1

meissadia:
\+1

Finding it impossible to get my long_desc displayed correctly. Line-wraps are sporadic once I use \x5.

For my sanity, I've simply redefined Basic#print_wrapped in my project to output the long_desc as-is.
```
class Thor
  module Shell
    class Basic
      def print_wrapped(message, options = {})
         stdout.puts message
       end
[...]

PR #739, opened 2020-10-18
🌧️ long_desc: new option to disable wrapping

igneus:
# What it does

The PR implements [this suggestion](https://github.com/erikhuda/thor/issues/398#issuecomment-364503506) from the #398 comments. `long_desc` now receives a new option `wrap`. Setting it to `false` disables wrapping of the long description and prints it verbatim.

```ruby
long_desc <<~MSG, wrap: false
  formatted text that is not wrapped
 ...
MSG
```

# What it breaks

It adds a new obligatory constructor argument to `Thor::Command`. If the constructor is considered part of the public API, this may be a problem.

*(I did read CONTRIBUTING.md, but I prefer not to use the rainbow emoji because of the meanings it usually carries nowadays, hence the cloud in PR title.)*

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -81,7 +81,7 @@ class Thor
         command.long_description = long_description if long_description
       else
         @long_desc = long_description
-        @long_desc_wrap = options[:wrap] || options[:wrap].nil?
+        @long_desc_wrap = options[:wrap] != false
       end
     end
 
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -65,8 +65,15 @@ class Thor
 
     # Defines the long description of the next command.
     #
+    # Long description is by default indented, line-wrapped and repeated whitespace merged.
+    # In order to print long description verbatim, with indentation and spacing exactly
+    # as found in the code, use the +wrap+ option
+    #
+    #   long_desc 'your very long description', wrap: false
+    #
     # ==== Parameters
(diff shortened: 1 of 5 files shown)


igneus:
If the change of `Thor::Command` constructor argument count were undesired, an alternative proposal would be to preserve the original argument count, but change the type of `long_description` from `String` to `String|Hash`, where the Hash form would allow passing presentation settings alongside text in a single argument. Like 

```ruby
Thor::Command.new('name', 'Description', {text: 'Long description ...', wrap: false}, nil, nil)
```

This would be 100% backward-compatible and would allow eventually adding more `long_desc` presentation settings in future.

lib/thor.rb
```diff
@@ -66,15 +66,23 @@ def desc(usage, description, options = {})
...
     # long description<String>
+    # options<Hash>
     #
     def long_desc(long_description, options = {})
       if options[:for]
         command = find_and_refresh_command(options[:for])
         command.long_description = long_description if long_description
       else
         @long_desc = long_description
+        @long_desc_wrap = options[:wrap] || options[:wrap].nil?
```
rafaelfranca, line 85:
```suggestion
        @long_desc_wrap = options[:wrap] != false
```

igneus: I agree that the proposed change makes the code look cleaner and easier to read.

PR #741, opened 2020-12-22
Use URI.send in apply action to work in Ruby 3

jaredcwhite:
In Ruby 3.0.0-rc1, I get the error `No such file or directory @ rb_sysopen` when using the `apply` action with a remote URL. By changing the use of `open` to `URI.send(:open …)`, it works in Ruby 3 while preserving backwards-compatibility (this is already how it's done in the `get` file manipulation method).

### DIFF
--- a/lib/thor/actions.rb
+++ b/lib/thor/actions.rb
@@ -219,7 +219,7 @@ class Thor
 
       contents = if is_uri
         require "open-uri"
-        open(path, "Accept" => "application/x-thor-template", &:read)
+        URI.send(:open, path, "Accept" => "application/x-thor-template", &:read)
       else
         open(path, &:read)
       end


jaredcwhite:
Looks like there's an issue with request stubs now in the test suite after making this change. I'd appreciate any pointers in how to resolve those.

excid3:
Ran into this today as well. 👍

lib/thor/actions.rb
```diff
@@ -219,7 +219,7 @@ def apply(path, config = {})
 
       contents = if is_uri
         require "open-uri"
-        open(path, "Accept" => "application/x-thor-template", &:read)
+        URI.send(:open, path, "Accept" => "application/x-thor-template", &:read)
```
excid3, line 222: Just curious, is `URI.open(path, ...)` not backwards compatible?

jaredcwhite: There's another part of the codebase that uses the same `send` pattern￼…I think because `open` used to be a private method.￼

excid3: TIL 👍

excid3:
@jaredcwhite looks like the webmock responses need an update to make the tests pass. Then hopefully we can get his merged!

PR #745, opened 2021-01-08
Add force option to the gsub_file action

AMHOL:
This allows for running gsub_file regardless of whether running in invoke or revoke mode


AMHOL:
Force pushed to add doc comment

PR #746, opened 2021-01-18
🌈 Replace `set` usages with arrays

deivid-rodriguez:
Set has been gemified as of ruby 3.0.0. Since `bundler` vendors `thor` internally, it's helpful for `bundler` that `thor` doesn't use `set`, so that `bundler` doesn't activate the gem `set` too early and final users can still specify whichever version of `set` they want in their Gemfiles.

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -1,4 +1,3 @@
-require "set"
 require_relative "thor/base"
 
 class Thor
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I took the chance to make sure that bundler works with all recent changes in thor's master branch. All good :+1:.

deivid-rodriguez:
Thank you :heart:

PR #747, opened 2021-02-03
Loaded the directory under tasks.

Mitsuru53:
### What I did
This PR is to fix the directory that thor goes to look for

When an arbitrary directory is created under the tasks directory, the thor files in that directory will also be read.

### DIFF
--- a/lib/thor/util.rb
+++ b/lib/thor/util.rb
@@ -211,7 +211,7 @@ class Thor
       #
       def globs_for(path)
         path = escape_globs(path)
-        ["#{path}/Thorfile", "#{path}/*.thor", "#{path}/tasks/*.thor", "#{path}/lib/tasks/*.thor"]
+        ["#{path}/Thorfile", "#{path}/*.thor", "#{path}/tasks/*.thor", "#{path}/lib/tasks/*.thor", "#{path}/lib/tasks/**/*.thor"]
       end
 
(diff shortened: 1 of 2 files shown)


lib/thor/util.rb
```diff
@@ -211,7 +211,7 @@ def thor_root_glob
       #
       def globs_for(path)
         path = escape_globs(path)
-        ["#{path}/Thorfile", "#{path}/*.thor", "#{path}/tasks/*.thor", "#{path}/lib/tasks/*.thor"]
+        ["#{path}/Thorfile", "#{path}/*.thor", "#{path}/tasks/*.thor", "#{path}/lib/tasks/*.thor", "#{path}/lib/tasks/**/*.thor"]
```
rafaelfranca, line 214: Does this also work?
```suggestion
        ["#{path}/Thorfile", "#{path}/*.thor", "#{path}/tasks/*.thor", "#{path}/lib/tasks/**/*.thor"]
```

Mitsuru53: @rafaelfranca 
Thank you for the review!
Yes, that works too! My code was redundant, I fix it.

Mitsuru53: @rafaelfranca 
Fixed.
I have confirmed that the specs all passed.
Please review again 😌

PR #752, opened 2021-04-13
🌈The klass parameter 'inject_into_class' should be given a string type.(also inject_into_module)

ratovia:
I think that in most cases the target klass is not defined in a file that uses inject_into_class.
So it is more friendly to set the default comment to string type.
as-is
```
inject_into_class "xxx.rb", ApplicationController, "Lorem"
```

to-be
```
inject_into_class "xxx.rb", "ApplicationController", "Lorem"
```

I think it's more realistic for test specs to delete the empty class definition in the first line.
Also about the inject_into_module.

### DIFF
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -210,9 +210,9 @@ class Thor
     #
     # ==== Examples
     #
-    #   inject_into_class "app/controllers/application_controller.rb", ApplicationController, "  filter_parameter :password\n"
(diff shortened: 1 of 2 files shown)


xjunior:
Looks good, but I think it's better if both are allowed. Since the implementation already allows that, you could just add another test to make sure strings are also supported. And +1 to update the docs.

rafaelfranca:
That works by "mistake" today and I prefer if we don't advertise it. I'll merge this as-is

Issue #627, opened 2018-10-09
Doubly nested subcommand creates wrong help output


Gargron:
Ref tootsuite/mastodon#8920

```ruby
module Mastodon
  class CLI < Thor
    desc 'settings SUBCOMMAND ...ARGS', 'Manage dynamic settings'
    subcommand 'settings', Mastodon::SettingsCLI
  end

  class RegistrationsCLI < Thor
    desc 'open', 'Open registrations'
    def open
    end

    desc 'close', 'Close registrations'
    def close
    end
  end

  class SettingsCLI < Thor
    desc 'registrations SUBCOMMAND ...ARGS', 'Manage state of registrations'
    subcommand 'registrations', RegistrationsCLI
  end
end
```

Expected output of `settings help` would point to `settings registrations open`, but instead only points to `registrations open`, which cannot be executed.

hostingnuggets:
Same issue here... or is nested subcommands not supported? I did not find any documentation either in the wiki mentioning nested subcommands.

jeffreycoe:
Just came across this issue, too. The link below has some additional details on the behavior we found: https://stackoverflow.com/questions/63331208/creating-nested-subcommands-using-thor/63341356

gangelo:
ATM, I simply override the banned class method in my "doubly nested subcommand" class:

```ruby
# "doubly nested subcommand"
...
class << self
  def ancestor_name
    :subcommnd_ancestor
  end

  def banner(command, namespace = nil, subcommand = false)
    command.formatted_usage(self, $thor_runner, subcommand).split("\n").map do |formatted_usage|
      "#{basename} #{ancestor_name} #{formatted_usage}"
    end.join("\n")
  end
end
...
```

PR #755, opened 2021-09-16
Fix for [Doubly nested subcommand creates wrong help output]

KishiTheMechanic:
Related this issue https://github.com/rails/thor/issues/627.
If we don't put command name (basename) there, it makes sense as arguments after the command.
Since it seems to take time to get subcommand names there, I think it's the solution for now.


dorner:
@KishiTheMechanic can you add a test for this behavior so we can validate it works as expected?

KishiTheMechanic:
I made #756 and close #755 now.

n1xn:
Little bit late on that, but I monkeypatched that [here](https://stackoverflow.com/a/77699855/3698577) if somebody else has that issue.

PR #759, opened 2021-11-04
Replace erikhuda/thor with rails/thor

koic:
This PR replaces erikhuda/thor with rails/thor. It will clarify the repository location and reduce redundant http redirects.


dorner:
Nice, me likes! @rafaelfranca

PR #761, opened 2021-12-22
Support latest did_you_mean

deivid-rodriguez:
Using `DidYouMean::SPELL_CHECKERS.merge!` has been deprecated.

Since this new version will be included with Ruby 3.1, I'm looking into supporting it without warnings.

PR #762, opened 2021-12-22
Use DidYouMean.correct_error for better Ractor support

yuki24:
I was working on better Ractor support for DidYouMean for the last few days. As part of that I had to deprecate the direct access to `DidYouMean::SPELL_CHECKERS` due to the necessity to maintain shareable objects within the dependency. From the next version the `DidYouMean.correct_error` method should be used as a replacement. This PR replaces the old API with the new one and should fix [the build failures](https://github.com/ruby/ruby/runs/4606737424?check_suite_focus=true#step:16:5131) in the upstream Ruby.

Let me know if there is anything I can do to get this change in.

### DIFF
--- a/lib/thor/error.rb
+++ b/lib/thor/error.rb
@@ -102,9 +102,14 @@ class Thor
   end
 
   if Correctable
-    DidYouMean::SPELL_CHECKERS.merge!(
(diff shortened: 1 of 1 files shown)


yuki24:
Ah, this is the exact same change as #761. Closing.

Issue #596, opened 2018-03-28
Difference between method_options and options


revolter:
http://whatisthor.com doesn't mention `method_options` at all, and https://github.com/erikhuda/thor/wiki/Method-Options doesn't mention `options` at all, hence, is very unclear what is the difference between them, as it looks like they do the same thing.

MaxLap:
`options` is an alias for `method_options`. You are right that the documentation could benefit from being updated.

azzenabidi:
@revolter Thanks for bringing this up! I just updated the wiki here: https://github.com/rails/thor/wiki/Method-Options @MaxLap I would love to update whatisthor.com too. Could you please point to the hompage source? cc @wifelette

MaxLap:
I just contributed a couple of patches to thor a couple of years ago. I have no idea where the source for that page is. Good luck!

azzenabidi:
@MaxLap Much appreciated, thanks!

azzenabidi:
Just raised a pull request #764 So hopefully, the issue is fixed.

PR #764, opened 2022-01-06
Added options' alias note 

azzenabidi:
I have updated both the wiki and the website to explain options and method_options. Original issue: https://github.com/rails/thor/issues/596

### DIFF
--- a/website/content/index.markdown
+++ b/website/content/index.markdown
@@ -166,6 +166,9 @@ files to keep your CLI description short and readable. You can then use
 Thor makes it easy to specify options and flags as metadata about a Thor
 command:
 
+ # Note  
+  option(and options)  are aliases for method_option (and methodo_options) respectively.
+
 <ruby>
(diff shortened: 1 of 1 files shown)


website/content/index.markdown
```diff
@@ -166,6 +166,9 @@ files to keep your CLI description short and readable. You can then use
 Thor makes it easy to specify options and flags as metadata about a Thor
 command:
 
+ # Note  
+  option(and options)  are aliases for method_option (and methodo_options) respectively.
```
dorner, line 170: There's a typo above. And I'd probably switch this around and say that `method_option` is an alias for `option` since the latter is much more commonly used.

azzenabidi: @dorner Thanks for the catch! fixed.

dorner (commented):
Looks good to me - @rafaelfranca ?

PR #767, opened 2022-01-10
Fix CI setting for Ruby 3.0

ytkg:
Due to actions/runner#849, we have to use quotes for 3.0.
Currently CI runs against latest 3 series, 3.1.


dorner:
LGTM, @rafaelfranca ?

PR #769, opened 2022-01-17
Repeatable hash type flags bugfix.

### LINKED ISSUE #768: Hash type is not supported by the repeatable option.
Here is an example to replicate the bug, as well as examples of running it, including the output and expected output. [Repeatable options](https://github.com/rails/thor/wiki/Method-Options#repeatable-options) don't seem to work with the hash type.

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require "thor"

class ExampleCLI < Thor
  desc "example", "shows how broken things are"
  option :doesnt,
    :required => false,
    :default => {},
    :repeatable => true,
    :type => :hash,
    :desc => "Should be repeatable but isn't"
  option :works,
    :repeatable => true,
    :required => false,
    :type => :string
  def create()
    puts options
  end

  def self.exit_on_failure?
    true
  end
end

ExampleCLI.start(ARGV)
```

You hand in two hashes and expect them to be set as two hashes in an array. What do you get? one hash, not in an array, with all key/vals merged into one hash.
```
./script/example create --works string1 --works string2 --doesnt key1:value1 key2:value2 --doesnt key3:val3 key4:val4
{"doesnt"=>{"key1"=>"value1", "key2"=>"value2", "key3"=>"val3", "key4"=>"val4"}, "works"=>["string1", "string2"]}
```

Instead, you should end up with an array of two different hashes:
```
{"doesnt"=>[{"key1"=>"value1", "key2"=>"value2"}, {"key3"=>"val3", "key4"=>"val4"}], "works"=>["string1", "string2"]}
```

Ordering doesn't seem to matter. It's still broken.
```
 ./script/example create --doesnt key3:val3 key4:val4 --works string1 --works string2 --doesnt key1:value1 key2:value2
{"doesnt"=>{"key3"=>"val3", "key4"=>"val4", "key1"=>"value1", "key2"=>"value2"}, "works"=>["string1", "string2"]}
```

if your keys in your objects are the same, youll end up with one object whos key values are the last ones you handed in.
```
./script/example create --works string1 --works string2 --doesnt key1:value1 key2:value2 --doesnt key1:dup1 key2:dup3
{"doesnt"=>{"key1"=>"dup1", "key2"=>"dup3"}, "works"=>["string1", "string2"]}
```

Instead, you should end up with an array of two different hashes:
```
{"doesnt"=>[{"key1"=>"value1", "key2"=>"value2"}, {"key1"=>"dup1", "key2"=>"dup3"}], "works"=>["string1", "string2"]}
```

patrick-motard:
They treated like every other type now. Instead of merging each key/val
into a hash, each flag is treated like it's own hash that should be put
into an array.

fixes #768

:rainbow:

### DIFF
--- a/lib/thor/parser/options.rb
+++ b/lib/thor/parser/options.rb
@@ -149,9 +149,7 @@ class Thor
   protected
 
     def assign_result!(option, result)
-      if option.repeatable && option.type == :hash
-        (@assigns[option.human_name] ||= {}).merge!(result)
-      elsif option.repeatable
+      if option.repeatable
         (@assigns[option.human_name] ||= []) << result
       else
         @assigns[option.human_name] = result
--- a/spec/parser/options_spec.rb
+++ b/spec/parser/options_spec.rb
@@ -417,7 +417,7 @@ describe Thor::Options do
 
       it "allows multiple values if repeatable is specified" do
         create :attributes => Thor::Option.new("attributes", :type => :hash, :repeatable => true)
-        expect(parse("--attributes", "name:one", "foo:1", "--attributes", "name:two", "bar:2")["attributes"]).to eq({"name"=>"two", "foo"=>"1", "bar" => "2"})
+        expect(parse("--attributes", "name:one", "foo:1", "--attributes", "name:two", "bar:2")["attributes"]).to eq([{"name"=>"one", "foo"=>"1"}, {"name"=>"two", "bar" => "2"}])
       end
     end
 


dorner:
Nice! Looks good to me - @rafaelfranca ?

wwwbruno:
Any updates on this fix? I faced this issue and this solves my problem.

rafaelfranca:
This is a change in the behavior and people might be expecting the option to merge instead of having two hashes. I don't think I can accept this change as is. I totally get the current behavior isn't what you want, but we can't assume it is broken and don't have a way to have it.

I'd accept an option that would allow the behavior you want though.

rafaelfranca (changes requested):
See above comment

PR #771, opened 2022-01-24
Fix matching when DidYouMean is enabled

petergoldstein:
Because of the changes in Ruby 3.1 to enable DidYouMean for a subset of errors, in a way that is somewhat confusing in specs, rspec-expectations now matches against the `original_message`, if the Error responds to that method.  That avoids having to match the backtrace in these circumstances.

Unfortunately that causes specs like this one - where DidYouMean is being used intentionally - to fail the matcher.  This PR updates the spec to avoid use of the matcher.

With this change, specs are now green.


dorner:
LGTM - @rafaelfranca

deivid-rodriguez:
This one can be closed now because a similar fix was introduced by #780.

PR #777, opened 2022-03-04
fix options spec.

timdiggins:
🌈  fix spec/parser/options_spec.rb:119 only (there are other CI failures on some rubies - will do independent PRs for those to fix)


voxik:
I wonder, shouldn't rather be the Thor::Correctable line above dropped instead? But that would probably not be backward compatible with older rspec-mock thinking

This is caused by https://github.com/rspec/rspec-expectations/pull/1339 just for the context.

timdiggins:

PR #778, opened 2022-03-04
try coveralls_reborn to fix ssl errors.

timdiggins:
Note that we could also use the coveralls action as recommended
in https://github.com/tagliala/coveralls-ruby-reborn
but it seems like a github token is needed, which makes
it more complex for contributors

This does mean dropping coveralls for EOLed rubies but
do we really need to post to coveralls on each test run?
Wouldn't one test run be enough?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ end
 
 group :test do
   gem "childlabor"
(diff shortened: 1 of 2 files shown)


timdiggins:

PR #779, opened 2022-03-04
fix specs for ruby 3 treatment of hash arg

timdiggins:
Just fixes ruby 3.0+ failures with specs expecting intracoversion between final hash argument and keyword args.

### DIFF
--- a/lib/thor/line_editor/readline.rb
+++ b/lib/thor/line_editor/readline.rb
@@ -1,12 +1,12 @@
+begin
+  require "readline"
+rescue LoadError
+end
+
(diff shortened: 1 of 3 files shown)


voxik:
Just FTR, this is not Ruby 3.x, but https://github.com/rspec/rspec-mocks/pull/1394 / https://github.com/rspec/rspec-mocks/issues/1460

timdiggins:
@voxik Ah! thanks - makes sense - feel free to rename to make more sense.

timdiggins:

PR #780, opened 2022-03-04
Fix CI to be green

timdiggins:
🌈  This is a combination of #777, #778, #779 if you prefer a single green PR 😀

### DIFF
--- a/spec/parser/options_spec.rb
+++ b/spec/parser/options_spec.rb
@@ -116,7 +116,9 @@ describe Thor::Options do
       expected = "Unknown switches \"--baz\""
       expected << "\nDid you mean?  \"--bar\"" if Thor::Correctable
 
-      expect { check_unknown! }.to raise_error(Thor::UnknownArgumentError, expected)
+      expect { check_unknown! }.to raise_error(Thor::UnknownArgumentError) do |error|
+        expect(error.to_s).to eq(expected)
+      end
     end
 
     it "skips leading non-switches" do
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ end
 
 group :test do
   gem "childlabor"
-  gem "coveralls", ">= 0.8.19"
+  gem 'coveralls_reborn', '~> 0.23.1', require: false if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.6.0")
   gem "rspec", ">= 3.2"
   gem "rspec-mocks", ">= 3"
   gem "rubocop", "~> 0.50.0"
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -1,13 +1,15 @@
 $TESTING = true
 
-require "simplecov"
-require "coveralls"
+if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.6.0")
+  require "simplecov"
+  require "coveralls"
 
-SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCov::Formatter]
+  SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCov::Formatter]
 
-SimpleCov.start do
-  add_filter "/spec"
-  minimum_coverage(90)
+  SimpleCov.start do
+    add_filter "/spec"
+    minimum_coverage(90)
+  end
 end
 
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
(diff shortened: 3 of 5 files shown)


dorner:
These all look good to me! @rafaelfranca

spec/parser/options_spec.rb
```diff
@@ -116,7 +116,9 @@ def remaining
       expected = "Unknown switches \"--baz\""
       expected << "\nDid you mean?  \"--bar\"" if Thor::Correctable
 
-      expect { check_unknown! }.to raise_error(Thor::UnknownArgumentError, expected)
+      expect { check_unknown! }.to raise_error(Thor::UnknownArgumentError) do |error|
+        expect(error.to_s).to eq(expected)
+      end
```
timdiggins, line 121: @voxik pointed out [elsewhere](https://github.com/rails/thor/pull/782#discussion_r823669341):

> I wonder, shouldn't rather be the `Thor::Correctable` line above dropped instead? But that would probably not be backward compatible with older rspec-mock 🤔

I can validate that removing that line instead does fix it for ruby 3.0.

> This is caused by [rspec/rspec-expectations#1339](https://github.com/rspec/rspec-expectations/pull/1339) just for the context.

But I don't think this is the same issue addressed by that PR, because https://github.com/rspec/rspec-expectations/pull/1339 was [included in  rspec-expectations v3.11.0](https://github.com/rspec/rspec-expectations/commit/44f9a2448c53d71eaea1f45ecb3473048bacb0de), and using that version, you still get an error in main.

I'm neutral on which way to fix it (just want the CI to be green 😀 )

voxik: Reported as https://github.com/rspec/rspec-expectations/issues/1362

timdiggins: @voxik can you clarify for me - is there any action needed on this PR or is it good to go in your opinion?

timdiggins:
@dorner Anything I can do to help get this merged?

dorner:
@rafaelfranca ping?

lib/thor/line_editor/readline.rb
```diff
@@ -1,12 +1,12 @@
...
+  require "readline"
+rescue LoadError
+end
+
 class Thor
   module LineEditor
     class Readline < Basic
       def self.available?
-        begin
-          require "readline"
```
rafaelfranca, line 6: @deivid-rodriguez wasn't this added because of bundler? if we move out to the top of the file would not that mean that bundler will require readline too early?

timdiggins: This change isn't necessary to get CI green (it was only so I could run one spec independently). If it's blocking the merge I can just remove this commit from the PR.

deivid-rodriguez: Yes, I added this because of Bundler. I don't recall the specifics, but it was something related to Windows I believe. I'd prefer if this commit was removed, yes.

I did reproduce the failing spec in isolation, but that can be fixed in the spec itself without giving up on requiring deadline lazily:

```diff
diff --git a/spec/line_editor_spec.rb b/spec/line_editor_spec.rb
index f034ec8..9c6250e 100644
--- a/spec/line_editor_spec.rb
+++ b/spec/line_editor_spec.rb
@@ -1,8 +1,9 @@
 require "helper"
+require "readline"
 
 describe Thor::LineEditor, "on a system with Readline support" do
   before do
-    @original_readline = ::Readline if defined? ::Readline
+    @original_readline = ::Readline
     silence_warnings { ::Readline = double("Readline") }
   end
 
@@ -22,10 +23,8 @@ end
 
 describe Thor::LineEditor, "on a system without Readline support" do
   before do
-    if defined? ::Readline
-      @original_readline = ::Readline
-      Object.send(:remove_const, :Readline)
-    end
+    @original_readline = ::Readline
+    Object.send(:remove_const, :Readline)
   end
 
   after do
```

deivid-rodriguez (commented):
Thanks for your @timdiggins, looks good to me! ❤️

lib/thor/line_editor/readline.rb
deivid-rodriguez: Is the failure what's fixed by #771?

deivid-rodriguez: Ah, ok! In that case I think it makes sense to merge this, and I can work on the fix for the ruby-head issue separately.

deivid-rodriguez: Also fixed running line editor specs in isolation at https://github.com/rails/thor/pull/791!

PR #782, opened 2022-03-04
Respect implicit encoding of thorfiles

### LINKED ISSUE #776: thor files have default encoding of ASCII-8BIT, not UTF-8
If you don't specify an encoding a thor file will be read as ASCII-8BIT, (unlike a Ruby file which will be UTF-8)

If you specify an encoding of UTF-8 explicitly, it will be respected.

This is very surprising. Especially when added to rubocop defaults which will not allow an encoding of UTF-8 in a ruby file. (https://docs.rubocop.org/rubocop/1.25/cops_style.html#styleencoding)

Reproduction: 
https://github.com/timdiggins/test-thor-encoding

Probably due to using File.binread at https://github.com/rails/thor/blob/main/lib/thor/util.rb#L153 ?

Maybe related to #191.

timdiggins:
🌈  fixes #776


timdiggins:
PS this has the #780 PR in it - will need to rebase once it has been merged.

spec/parser/options_spec.rb
```diff
@@ -116,7 +116,9 @@ def remaining
       expected = "Unknown switches \"--baz\""
       expected << "\nDid you mean?  \"--bar\"" if Thor::Correctable
 
-      expect { check_unknown! }.to raise_error(Thor::UnknownArgumentError, expected)
```
voxik, line 119: I wonder, shouldn't be rather the `Thor::Correctable` line above dropped? But that would probably not be backward compatible with older rspec-mock :thinking:

This is caused by https://github.com/rspec/rspec-expectations/pull/1339 just for the context.

timdiggins: @voxik sorry for confusion, this comes from #780, because I wanted this to be green first before proposing a change.

Can we continue the discussion in #780 ?

timdiggins: https://github.com/rails/thor/pull/780/files#r823694117

voxik: Ah, sorry, right. I have noticed just now you opened multiple PRs and there is one for this case specifically.

voxik: Actually this comes from #777 :)

timdiggins:
@dorner @deivid-rodriguez I've rebased this on main, now it's (mostly) green

PR #784, opened 2022-04-19
Check if `type: array` values are in `enum`

### LINKED ISSUE #783: `enum` should be enforced for all values of `type: array`
🌈

`enum` was added in https://github.com/rails/thor/pull/237, but it originally only validated `type: string` values.

Validation of `type: numeric` values was added in https://github.com/rails/thor/pull/386

I expected all values of my `type: array` option to also be validated.

## Example

`cli.rb`:

```ruby
require "thor"

class MyCLI < Thor
  desc "array_with_enum [--fruits=apple banana]", "Demo of using an enum with an array"
  option :fruits, type: :array, enum: ["apple", "banana", "clementine"]
  def array_with_enum
    puts "fruits: #{options[:fruits]}"
  end
end

MyCLI.start(ARGV)
```

This works:

```
ruby ./cli.rb array_with_enum --fruits=banana apple
fruits: ["banana", "apple"]
```

This also works:

```
ruby ./cli.rb array_with_enum --fruits=kiwi
fruits: ["kiwi"]
```

but shouldn't since "kiwi" is not a fruit in the `enum`.

movermeyer:
🌈

### What are you trying to accomplish?

Fixes #783 

### What approach did you choose and why?

Extending the pattern done for `type: string` and `type: numeric`.

### What should reviewers focus on?

🤷 

### The impact of these changes

`enum` validation will be done for `type: array` option values.

### DIFF
--- a/lib/thor/parser/arguments.rb
+++ b/lib/thor/parser/arguments.rb
@@ -122,7 +122,15 @@ class Thor
     def parse_array(name)
       return shift if peek.is_a?(Array)
       array = []
-      array << shift while current_is_value?
+      while current_is_value?
+        value = shift
+        if !value.empty? && @switches.is_a?(Hash) && switch = @switches[name]
+          if switch.enum && !switch.enum.include?(value)
+            raise MalformattedArgumentError, "Expected all values of '#{name}' to be one of #{switch.enum.join(', ')}; got #{value}"
+          end
+        end
+        array << value
+      end
       array
     end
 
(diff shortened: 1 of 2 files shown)


lib/thor/parser/arguments.rb
```diff
@@ -122,7 +122,15 @@ def parse_hash(name)
     def parse_array(name)
       return shift if peek.is_a?(Array)
       array = []
-      array << shift while current_is_value?
+      while current_is_value?
+        value = shift
+        if !value.empty? && @switches.is_a?(Hash) && switch = @switches[name]
```
dorner, line 127: Can we restructure this conditional a bit? The `=` towards the end is kind of like a shark in calm waters, it's really unexpected.

PR #785, opened 2022-05-03
Switch hash from MD5 to SHA256

### LINKED ISSUE #287: FIPS compatibility
I am testing Thor in FIPS mode [1]. Unfortunately it fails with following error:

```
md5_dgst.c(78): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
```

Would you please consider to use different hashing algorithm such as SHA256 to make Thor FIPS compatible?

[1] http://en.wikipedia.org/wiki/FIPS_140

stanhu:
On FIPS-compliant systems (http://en.wikipedia.org/wiki/FIPS_140), MD5
cannot be used. Switch to SHA256 instead.

However, this change does not keep backward compatibility with systems
with already-installed Thor recipes.

Closes #287

### DIFF
--- a/lib/thor/runner.rb
+++ b/lib/thor/runner.rb
@@ -2,7 +2,7 @@ require_relative "../thor"
 require_relative "group"
 
 require "yaml"
-require "digest/md5"
+require "digest/sha2"
 require "pathname"
 
 class Thor::Runner < Thor #:nodoc: # rubocop:disable ClassLength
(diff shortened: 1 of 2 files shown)


stanhu:
Some tests will be fixed by https://github.com/rails/thor/pull/771.

stanhu:
@rafaelfranca Could I trouble you to release a new version with this? Thanks so much.

PR #786, opened 2022-05-18
CreateFile#identical? fixed for files containing multi-byte UTF-8 codepoints

tomclose:
🌈

Currently if you regenerate a file that contains a multi-byte UTF-8 codepoint (such as €) the file will falsely be identified as having changed. This PR adds a test to the create_file_spec to identify the problem, and then fixes the problem.

The change is the same as in https://github.com/rails/thor/pull/656, but that PR seems to have languished unmerged for just over 3yrs. I'm hoping the testing approach in the PR will address the previous objections so that the change can be merged.

### DIFF
--- a/lib/thor/actions/create_file.rb
+++ b/lib/thor/actions/create_file.rb
@@ -43,7 +43,7 @@ class Thor
       # Boolean:: true if it is identical, false otherwise.
       #
       def identical?
-        exists? && File.binread(destination) == render
+        exists? && File.binread(destination) == String.new(render).force_encoding("ASCII-8BIT")
       end
 
       # Holds the content to be added to the file.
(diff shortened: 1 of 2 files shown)


lib/thor/actions/create_file.rb
```diff
@@ -43,7 +43,7 @@ def initialize(base, destination, data, config = {})
       # Boolean:: true if it is identical, false otherwise.
       #
       def identical?
-        exists? && File.binread(destination) == render
+        exists? && File.binread(destination) == String.new(render).force_encoding("ASCII-8BIT")
```
dorner, line 46: I'd add a comment here indicating that `binread` defaults to this encoding to explain why this works.

dorner (commented):
Added one comment. Otherwise I like this!

PR #788, opened 2022-06-03
Fix specs to work with latest rspec-mock

deivid-rodriguez:
It's more strict regarding mocking keywords args vs positional args.

This is currently making CI fail. I didn't notice it locally at #787, because I had an older `Gemfile.lock` file locally that was locked to an older version of `rspec-mocks` that didn't have the issue.


deivid-rodriguez:
Actually I just noticed there's a previous more complete PR proposing the same thing, so I'll close this!

PR #789, opened 2022-06-16
Add support for new `Exception#detailed_message` behaviour

deivid-rodriguez:
Ruby 3.2 will introduce `Exception#detailed_message` and `did_you_mean` has been already updated in Ruby 3.2 to use that.

The new behaviour means not changing the original `Exception#message`. That means it is hard to get the previous error output, because `Exception#detailed_message` includes not only `did_you_mean` decorations, but also additional information like the exception class.

There's two cases where the did_you_mean message is displayed, depending on whether we are in debug mode or not.

* In debug mode, we let the original exception be raised. In this case, I think it's fine to let the new wording blow up, so I just changed the assertions on the Exception raised to comply with the new behaviour.

* In non debug mode, we print a custom error. In this case, I slightly prefer the previous Thor output better than the one that `Exception#detailed_message` gives. The current one would be

  > Could not find command "paintz" in "barn" namespace.
  > Did you mean?  "paint"

  vs the new one

  > Could not find command "paintz" in "barn" namespace. (Thor::UndefinedCommandError)
  > Did you mean?  "paint"

  So I changed `Thor::Base` to manually build the message so that we keep
the same error wording.


deivid-rodriguez:
Marked as draft because I want to find a better way.

deivid-rodriguez:
Ok, I think the new approach is much cleaner, this should be ready!

PR #791, opened 2022-06-16
Allow running readline specs in isolation

deivid-rodriguez:
Plus other minor improvements since I was at it.


timdiggins (commented):
just as a double check, I've validated on my machine that this does the trick 

(on main, I get an error when I run `bundle exec rspec spec/line_editor_spec.rb`, whereas on this branch it succeeds).

PR #796, opened 2022-08-07
Respect the updated NO_COLOR specification

### LINKED ISSUE #795: Update thor's implementation of the NO_COLOR spec to match the updated specification.
As discussed in [this comment thread](https://github.com/rails/thor/commit/21e65684432c0fcc975c1d6fda497da24149f763#r80414525) on the original commit the NO_COLOR specification has been silently changed. The new specification expects that NO_COLOR will only be respected if set to a non-empty value.

Given the arguments in the NO_COLOR project discussion it's probably worth updating it here.

I'll submit a PR for this issue shortly.

coderjoe:
Since implementation the NO_COLOR specification has changed. Previously
if NO_COLOR was set in the environment it should be respected regardless
of value. In the new updated specification NO_COLOR should only be
respected if set to any non-empty string.

See this comment thread on the original commit for more information.
https://github.com/rails/thor/commit/21e65684432c0fcc975c1d6fda497da24149f763#r80414525

Fixes #795

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,8 @@ group :test do
   gem "rspec", ">= 3.2"
   gem "rspec-mocks", ">= 3"
   gem "simplecov", ">= 0.13"
-  gem "webmock"
+  gem "webmock", '~> 3.14.0' if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.4.0")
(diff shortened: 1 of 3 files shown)


rafaelfranca:
I think it is fine to merge this. I'll make sure this is included in the changelog.

coderjoe:
Thank you very much for your help @rafaelfranca :)

PR #797, opened 2022-10-26
Don't document negative boolean option named `no_*`

BrentWheeldon:
🌈

I ran into this earlier today where I had underscored option names. Let me know if there is anything else I should include! 🙇🏻


dorner:
@BrentWheeldon looks like this fails on some versions of Ruby!

BrentWheeldon:
@dorner Ah - I didn't realise that `String#starts_with?` only relatively recently started accepting a regex. Fixed!

dorner:
Nice, looks good to me! @rafaelfranca

PR #800, opened 2022-11-14
Fix dashless option usage info

sambostock:
🌈

When option aliases are specified without dashes

```ruby
option :from, aliases: [:f]
```

they are correctly handled when parsing the command

```
my_cli hello -f Loki Thor
from: Loki
Hello Thor
```

However the usage information does not include the dash. This PR fixes that by moving alias normalization to happen sooner, so it applies to both the usage information generator and the actual command:

```diff
 my_cli help hello  
 Usage:
   my_cli hello NAME

 Options:
-  f, [--from=FROM]  
+  -f, [--from=FROM]  

 say hello to NAME
```

### DIFF
--- a/spec/parser/option_spec.rb
+++ b/spec/parser/option_spec.rb
@@ -108,8 +108,8 @@ describe Thor::Option do
         expect(parse([:foo, :b, "--bar"], true).name).to eq("foo")
       end
 
-      it "sets all other items as aliases" do
-        expect(parse([:foo, :b, "--bar"], true).aliases).to eq([:b, "--bar"])
+      it "sets all other items as normalized aliases" do
+        expect(parse([:foo, :b, "--bar"], true).aliases).to eq(["-b", "--bar"])
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/parser/option_spec.rb
```diff
@@ -105,11 +105,11 @@ def option(name, options = {})
 
     describe "with key as an array" do
       it "sets the first items in the array to the name" do
-        expect(parse([:foo, :bar, :baz], true).name).to eq("foo")
+        expect(parse([:foo, :b, "--bar"], true).name).to eq("foo")
```
sambostock, line 108: As far as I can tell, these two tests were invalid.

Specifying a multicharacter alias without including leading dashes leads to a single dash being prepended (rather than two), which is invalid, as there is no way to invoke it.

By example:

Specified Alias|Final Alias
-|-
`"-b"`, `:"-b"`, `"b"`, `:b`,|`-b`
`"--bar"`, `:"--bar"`|`--bar`
`"bar"`, `:bar`|`-bar` (invalid)

Therefore, I updated these tests to use legitimate aliases, which then make sense with normalization.

Specifying invalid aliases is still allowed, it just now will generate a single dash in the usage documentation (which reflects what is actually going on).

Another PR could address forbidding invalid aliases (as they do not work, and would be misleading) in isolation.

PR #802, opened 2022-12-13
Prefer to use URI.open and File.open instead of Kernel.open

hsbt:
👋  CodeQL detects "Use of `Kernel.open` or `IO.read` with a non-constant value" warnings in https://github.com/ruby/ruby. I hope to suppress them.

### DIFF
--- a/lib/thor/actions.rb
+++ b/lib/thor/actions.rb
@@ -225,7 +225,7 @@ class Thor
         require "open-uri"
         URI.open(path, "Accept" => "application/x-thor-template", &:read)
       else
-        open(path, &:read)
+        File.open(path, &:read)
       end
 
       instance_eval(contents, path)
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I fixed the `URI.open` ones similarly at #787. Not yet attended unfortunately!

hsbt:
Oh! Thanks to share it.

deivid-rodriguez:
Should we enable [Security/Open](https://docs.rubocop.org/rubocop/1.40/cops_security.html#securityopen) in RuboCop configuration file too?

hsbt:
👍 or added CodeQL actions.

deivid-rodriguez:
That too, just thought of RuboCop since it's already used.

hsbt:
@rafaelfranca done 🚀

hsbt:
Thanks @rafaelfranca ❤️

PR #815, opened 2023-03-02
Allow force-enabling colors

john-h-k:
Useful for places such as GitHub actions where you want colour, but are not in a TTY


rafaelfranca:
Can you please add a test for this new environment variable and documentation?  Thank you

PR #820, opened 2023-05-05
Allow setting file permissions with create_file

skipkayhil:
Ref rails/rails#48135

Previously, to restrict the permissions of a file created using create_file, chmod could be called right after. However, that leaves a short amount of time between when the file is first created and when the permissions are updated where another user could read the file.

This commit enables passing file permissions to create_file so that permissions can be set when the file is created.


casperisfine:
LGTM (ci failures are unrelated).

@rafaelfranca do you wanna have a look? Otherwise I'll merge.

rafaelfranca:
👍🏽

Issue #819, opened 2023-04-06
Default value not printed in help if option type boolean and default value is false


hadmut:
Hi, 

usually the help output generated by a thor app lists the options for the class and the actions, and shows the default value if a default value is set. 

But for 
  method_option :save,  type: :boolean, desc: "save file" , default: false

the default value is printed only of it is true, not false.  (Thor 1.2.1)

regards

p8:
@hadmut this can now be closed, right?

hadmut:
I didn't find the time to verify it yet, but I do fully trust you.

PR #849, opened 2023-06-12
Print default in help when option type is :boolean and default is false

nevesenin:
🌈 I just learned, that there is an open issue #819 already, which will be solved by this PR.

### DIFF
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -105,7 +105,12 @@ class Thor
     end
 
     def show_default?
-      super || [TrueClass, FalseClass].any? { |c| default.is_a?(c) }
+      case default
+      when TrueClass, FalseClass
+        true
+      else
+        super
+      end
     end
 
     VALID_TYPES.each do |type|
--- a/lib/thor/parser/option.rb
+++ b/lib/thor/parser/option.rb
@@ -104,6 +104,10 @@ class Thor
       end
     end
 
+    def show_default?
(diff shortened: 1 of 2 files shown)


lib/thor/parser/option.rb
```diff
@@ -104,6 +104,10 @@ def aliases_for_usage
       end
     end
 
+    def show_default?
+      super || [TrueClass, FalseClass].any? { |c| default.is_a?(c) }
+    end
```
rafaelfranca, line 109:
```suggestion
    def show_default?
      case default
      when TrueClass, FalseClass
        true
      else
        super
      end
    end
```

spec/parser/option_spec.rb
```diff
@@ -269,4 +269,21 @@ def option(name, options = {})
       end
     end
   end
+
+  describe "#print_default" do
+    it "prints boolean with true default value" do
+      expect(option(:foo,    {
```
rafaelfranca, line 275:
```suggestion
      expect(option(:foo, {
```

spec/parser/option_spec.rb
```diff
@@ -269,4 +269,21 @@ def option(name, options = {})
...
+        required: false,
+        type: :boolean,
+        default: true
+      }).print_default).to eq(true)
+    end
+    it "prints boolean with false default value" do
+      expect(option(:foo,    {
```
rafaelfranca, line 282:
```suggestion
      expect(option(:foo, {
```

PR #854, opened 2023-08-09
Extract print methods to seperate classes

p8:
The printing methods are already pretty large, making it difficult to add extra functionality. Extracting them to seperate classes allows refactoring them for easier maintainability.

A lot of the functionality of calculating the terminal width can be extracted to a separate object as well.

### DIFF
--- a/lib/thor/shell/basic.rb
+++ b/lib/thor/shell/basic.rb
@@ -1,8 +1,10 @@
+require_relative "column_printer"
+require_relative "table_printer"
+require_relative "wrapped_printer"
+
 class Thor
   module Shell
     class Basic
-      DEFAULT_TERMINAL_WIDTH = 80
-
       attr_accessor :base
       attr_reader   :padding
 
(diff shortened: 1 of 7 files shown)


lib/thor/shell/printer.rb
```diff
@@ -0,0 +1,13 @@
+class Thor
+  module Shell
+    class Printer
```
rafaelfranca, line 3: Do we need this class? Can't we just merge it with ColumnPrinter?

p8: `TablePrinter` and `WrappedPrinter` inherit from `Printer` as well.
All three subclasses implement the `print` method in their own way.

rafaelfranca: Can't we change them to inherit from `ColumnPrinter`?

rafaelfranca: I'm asking this because Printer doesn't have the same interface as their subclasses so it is breaking the Liskov substitution principle, and I see no reason to break that.

p8: Done.

PR #865, opened 2023-12-14
Fix typo

takmar:


lib/thor/shell/basic.rb
```diff
@@ -143,14 +143,14 @@ def say_status(status, message, log_status = true)
         stdout.flush
       end
 
-      # Make a question the to user and returns true if the user replies "y" or
+      # Make a question to the user and returns true if the user replies "y" or
```
p8, line 146: While we are changing this we can improve the sentences a bot as well:
```suggestion
      # Asks the user a question and returns true if the user replies "y" or
```

lib/thor/shell/basic.rb
```diff
@@ -143,14 +143,14 @@ def say_status(status, message, log_status = true)
...
       # "yes".
       #
       def yes?(statement, color = nil)
         !!(ask(statement, color, add_to_history: false) =~ is?(:yes))
       end
 
-      # Make a question the to user and returns true if the user replies "n" or
+      # Make a question to the user and returns true if the user replies "n" or
```
p8, line 153:
```suggestion
      # Asks the user a question and returns true if the user replies "n" or
```

p8 (changes requested):

lib/thor/shell/basic.rb
takmar: Thank you for your review! I fixed it.

p8 (approved):

PR #877, opened 2024-02-23
feat: support `gsub_file` erroring if gsub doesn't change anything, and add `gsub_file!`

### LINKED ISSUE #874: Ability to have `gsub_file` error if it didn't gsub anything
[I recently discovered](https://github.com/ackama/rails-template/pull/533) that `gsub_file` does not error when it does not match; this feels like a bit of a footgun to me.

I've worked around this using this utility function to ensure the file contents has changed: 

```
def gsub_file!(path, flag, *args, &block)
  content = File.binread(path)

  gsub_file(path, flag, *args, &block)

  raise StandardError, "the contents of #{path} did not change!" if content == File.binread(path)
end
```

However, I think it would be a good addition to `thor` itself - I'm thinking maybe `gsub_file "", "", "", error_on_no_change: true` and then an alias of `gsub_file!`?

G-Rath:
🌈

---

This is a straw-person solution for #874  - overall I think it's not bad but expect changes to be requested; in particular, the option name is pretty verbose so happy if someone wants to suggest an alternative though to me really its about having `gsub_file!`.

A bonus to this is that anything using `gsub_file` also supports this behaviour such as `comment_lines` and `uncomment_lines` - I think it would be nice to have bang versions of such methods too but want to get the core behaviour/implementation nailed down first.

Resolves #874

### DIFF
--- a/lib/thor/actions/file_manipulation.rb
+++ b/lib/thor/actions/file_manipulation.rb
@@ -263,9 +263,12 @@ class Thor
     def gsub_file!(path, flag, *args, &block)
       config = args.last.is_a?(Hash) ? args.pop : {}
 
-      config[:error_on_no_change] = true
+      return unless behavior == :invoke || config.fetch(:force, false)
 
-      gsub_file(path, flag, *args, config, &block)
+      path = File.expand_path(path, destination_root)
+      say_status :gsub, relative_to_original_destination_root(path), config.fetch(:verbose, true)
+
+      actually_gsub_file(path, flag, args, true, &block) unless options[:pretend]
     end
 
     # Run a regular expression replacement on a file.
@@ -274,8 +277,7 @@ class Thor
     # path<String>:: path of the file to be changed
     # flag<Regexp|String>:: the regexp or string to be replaced
     # replacement<String>:: the replacement, can be also given as a block
-    # config<Hash>:: give :verbose => false to not log the status,
-    #                :error_on_no_change => true to raise an error if the file does not change, and
+    # config<Hash>:: give :verbose => false to not log the status, and
     #                :force => true, to force the replacement regardless of runner behavior.
     #
     # ==== Example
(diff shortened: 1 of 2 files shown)


lib/thor/actions/file_manipulation.rb
```diff
@@ -260,6 +261,32 @@ def inject_into_module(path, module_name, *args, &block)
     #     match << " no more. Use thor!"
     #   end
     #
+    def gsub_file!(path, flag, *args, &block)
+      config = args.last.is_a?(Hash) ? args.pop : {}
+
+      config[:error_on_no_change] = true
```
rafaelfranca, line 267: I'm ok with this feature but I don't think we should add a config to `gsub_file` to implement this. I think we can implement this by moving the logic to a private method that we can then add the option, or just use a block to change the behavior using the template method pattern.

rafaelfranca (changes requested):

lib/thor/actions/file_manipulation.rb
G-Rath: I've pushed up a change that is hopefully in the right direction - let me know what you think.

At this point I personally feel having the option on `gsub_file` is better because it's more flexible and feels like the DRYest way to do it already, but `gsub_file!` is most important to me so happy to go with your preference (and maybe I'm missing something about what you mean - I'm not sure what the "template method pattern" is at this point).

G-Rath:
@rafaelfranca could I get a re-review on this please?

G-Rath:
@rafaelfranca would you mind re-reviewing this? we'd really like to get it landed and I think I've addressed your concerns 🙂

G-Rath:
@rafaelfranca @yahonda would it be possible to have this reviewed, or at least have the `need-work` label removed?

rafaelfranca:
Yes! Happy to

G-Rath:
@rafaelfranca I've opened #908 adding a bang version of `insert_into_file` - I realized after doing that it doesn't actually make sense to have versions of `prepend_into_file` and `append_into_file` since they should never be able to fail unless the file doesn't exist or is not writable, which already causes an error

Issue #868, opened 2024-01-14
Commands that contain hyphens cannot be executed


naomichi-y:
In 1.3.0, it seems that the command cannot be executed if the command name includes hyphen.

```
class Test < Thor
  desc "foo-bar", "an example task"
  def foo_bar
    puts "I'm a thor task!"
  end
end
```

```
$ bundle exec thor test:foo-bar
Commands:
  thor test:foo-bar         # an example task
  thor test:help [COMMAND]  # Describe available commands or one specific command
```

Up to 1.2.2, the command was executed as intended.
```
$ bundle exec thor test:foo-bar
I'm a thor task!
```

Is this a feature change?
Looking at the change history, I can't find any related issues.

naomichi-y:
This is probably caused by `Thor::Base.subclasses.detect` not finding `foo-bar`.
By converting to `command.tr('-', '_')`, the behavior is the same as up to 1.2.2.

https://github.com/rails/thor/pull/754/files#diff-5393308169879c36b527bdc9a24a37e0e3939ad5e4a7b359d327645a7f5a714cR136

rafaelfranca:
Not intended. Can you please open a PR?

takmar:
@rafaelfranca @naomichi-y 
Hi. I fixed the issue so I created [PR](https://github.com/rails/thor/pull/878). 

In addition to this problem, it seems that the alias method also did not work properly.

PR #878, opened 2024-03-22
Correctly identify hyphenated and alias command names

takmar:
Fixes https://github.com/rails/thor/issues/868

Problem:
After version 1.3.0, invoking commands with hyphenated and alias names does not function as expected.

Given a Thor class definition as follows:
```
class Test < Thor
  desc "foo-bar", "an example task"
  def foo_bar
    puts "I'm a thor task!"
  end

  map "fb" => :foo_bar 
end
```

Each command results in the following.

```
% thor test:foo_bar
I'm a thor task!
```

```
% thor test:foo-bar
Commands:
  thor test:foo-bar         # an example task
  thor test:help [COMMAND]  # Describe available commands or one specific command
```

```
% thor test:fb
Commands:
  thor test:foo-bar         # an example task
  thor test:help [COMMAND]  # Describe available commands or one specific command
```

This PR addresses the issue by implementing a fix in the `find_class_and_command_by_namespace` method, ensuring that it correctly identifies both hyphenated command names and their aliases.

### DIFF
--- a/lib/thor.rb
+++ b/lib/thor.rb
@@ -439,6 +439,17 @@ class Thor
       command && disable_required_check.include?(command.name.to_sym)
     end
 
+    # Checks if a specified command exists.
+    #
+    # ==== Parameters
+    # command_name<String>:: The name of the command to check for existence.
+    #
+    # ==== Returns
+    # Boolean:: +true+ if the command exists, +false+ otherwise.
+    def command_exists?(command_name) #:nodoc:
+      commands.keys.include?(normalize_command_name(command_name))
+    end
+
   protected
 
     # Returns this class exclusive options array set.
--- a/lib/thor/group.rb
+++ b/lib/thor/group.rb
@@ -211,6 +211,17 @@ class Thor::Group
       raise error, msg
     end
 
+    # Checks if a specified command exists.
+    #
+    # ==== Parameters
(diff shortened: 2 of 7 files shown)


naomichi-y:
It worked correctly 👍

bastelfreak:
hey people, any chance this can get reviewed and merged? This fixes CI errors in one of my projects, https://github.com/voxpupuli/modulesync.

lib/thor.rb
```diff
@@ -439,6 +439,17 @@ def disable_required_check?(command) #:nodoc:
...
     end
 
+    # Checks if a specified command exists.
+    #
+    # ==== Parameters
+    # command_name<String>:: The name of the command to check for existence.
+    #
+    # ==== Returns
+    # Boolean:: +true+ if the command exists, +false+ otherwise.
+    def command_exists?(command_name)
```
rafaelfranca, line 449: Do we need this method in the public API? Should we `:nodoc:` it instead? I don't think we have a need to expose this to users.

takmar: Thank you for reviewing. I added `:nodoc:`.

bastelfreak:
@rafaelfranca thanks for the merge! Would it be possible to make a new release?

PR #892, opened 2024-12-04
Lazily load YAML

deivid-rodriguez:
This avoids errors when Bundler is activated _after_ thor, and a different version of YAML (or dependencies of it, like stringio) is included in the Gemfile/Gemfile.lock than the one that thor activates internally.


deivid-rodriguez:
Any feedback/concerns here?

PR #893, opened 2024-12-16
Use secure thor link

hlascelles:
The link to whatisthor.com is insecure: `http://whatisthor.com/` 

It should be the secure version.


hlascelles:
@rafaelfranca @yahonda I hope this is a straightforward one?

PR #898, opened 2025-03-10
Fix encoding error when trying to show a diff:

Edouard-chin:
Ref https://github.com/rails/rails/issues/54733

- ### Problem

  Copying or creating a file with UTF-8 charatacters would raise an encoding conversion error when:

  1) There is a collision and the user wants to see a diff.
  2) The application has set its `Encoding.default_internal` to a non-nil value.

  ### Context

  In Rails applications, the `Encoding.default_internal` is set to UTF-8. The `Encoding.default_external` will be used to transcode a string because the `default_internal` is not nil. So doing this would raise a `Encoding::UndefinedConversionError`

  ```ruby
  content = "\xE2\x80\x99".force_encoding(Encoding::ASCII_8_BIT)
  File.write("my_file.rb", content)
  ```

  In Thor's case, the encoding of the source file will always be `ASCII_8_BIT` because we are reading the file in binary mode: https://github.com/rails/thor/blob/3178667e1727504bf4fb693bf4ac74a5ca6c691e/lib/thor/actions/file_manipulation.rb#L26


skipkayhil (approved):
Love simple changes, makes sense to me 👍

PR #900, opened 2025-04-17
Use git difftool for merge.tool identifiers

moguls753:
As mentioned in rails/rails#54912, Thor’s `merge_tool` used to invoke the raw Git `merge.tool` value
directly, which fails for built‑in identifiers like vimdiff3.

This change updates `git_merge_tool` to:

1. Read any custom command from `mergetool.<tool>.cmd`
2. If the tool name resolves to a real executable in $PATH, use it
3. Otherwise fall back to `git difftool --no-index --tool=<tool>`

Fixes rails/rails#54912

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,13 +5,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.6', '2.7', '3.0', '3.1', '3.2', '3.3', 'head']
+        ruby: ['2.6', '2.7', '3.0', '3.1', '3.2', '3.3', '3.4','head']
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
     - uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
-        bundler: ${{ (matrix.ruby_version < '3' && '2.4.21') || 'latest' }}
+        bundler: ${{ (matrix.ruby < '3' && '2.4.21') || 'latest' }}
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - run: bundle exec thor spec
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,8 @@ group :test do
   gem "rspec-mocks", ">= 3"
   gem "simplecov", ">= 0.13"
   gem "webmock", ">= 3.14"
+  gem "rdoc"
+  gem "readline"
 end
 
 gemspec


lib/thor/shell/basic.rb
```diff
@@ -380,8 +380,18 @@ def merge_tool #:nodoc:
...
+      def git_merge_tool
+        tool = `git config merge.tool`.strip
+        return "" if tool.empty?
+
+        custom = `git config --get mergetool.#{tool}.cmd`.strip
+        return custom unless custom.empty?
+
+        return tool if system("command -v #{Shellwords.escape(tool)} > /dev/null 2>&1")
+
+        "git mergetool --no-prompt --tool=#{tool}"
```
Edouard-chin, line 392: Unfortunately I don't think this works. `git mergetool` will only work when there are conflicting files when attempting to merge (in the context of git). In thor's case there is no "conflict", so `git mergetool` will just bailout.

moguls753: Yes you're right. Maybe it would be better to use `git difftool --no-index` instead, since `git mergetool` is looking for conflicts in git's index. The manpage of `git difftool` says:

> git difftool falls back to git mergetool config variables when the difftool equivalents have not been defined.

If nothing is configured you'll get:

> This message is displayed because 'diff.tool' is not configured.
See 'git difftool --tool-help' or 'git help config' for more details.
'git difftool' will now attempt to use one of the following tools:
opendiff kdiff3 tkdiff xxdiff meld kompare gvimdiff diffuse diffmerge ecmerge p4merge araxis bc codecompare smerge nvimdiff vimdiff emerge
Viewing (1/1): '~/app/config/application.rb20250418-22187-1vvz3o.rb'
Launch 'vimdiff' [Y/n]? 

This seems like the best approach; adding --no-prompt will suppress the configuration message. What do you think?

Edouard-chin: You may be able to confirm but this doesn't work either. `git difftool` will create 2 **readonly** buffers. So users will be presented with the diff but won't be able to make any changes. Also I don't think `--no-index` is a thing on the `difftool` command.

I had a quick look to find a solution and this is way more tricky. Basically the problem is that running a `git` command is not suited for what we are trying to do. 
Our best bet if we wanted to stick with git and respect a user's configuration would be to create our own shell executable and use the `git mergetool-lib` https://git-scm.com/docs/git-mergetool--lib

But tbh this seems like a lot of work and it's probably best to ask a maintainer of this repo whether they'd be ok with such change.

moguls753: Thanks for your feedback! You're right `git difftool` opens 2 **readonly** buffers, but this is just a flag, you can in fact save your changes, either by using `:wq!` or by unsetting the flag via `:set noreadonly`. I think it just calls `vim -R` internally. From `:help readonly`:
> 'readonly' 'ro'		boolean	(default off)
			local to buffer  |local-noglobal|
	If on, writes fail unless you use a '!'.  Protects you from
	accidentally overwriting a file.  Default on when Vim is started
	in read-only mode ("vim -R") or when the executable is called "view".

Regarding `--no-index`: `git difftool` is just a frontend for `git diff`, it does support the `--no-index` flag as we need it in this context.

Edouard-chin: Thanks for the explanation! Great, then that make sense to me 👍 .

